PR #176, opened 2018-10-20
175 windows app fixes

KrisDavie:
This should fix the windows electron app.

@mase5 can you test on linux? Particularly `npm run electron` and the open loom folder button, I had to change that. Either the package updated, or it is a bit different on windows.


KrisDavie:
macOS is broken for some reason. Will wait to merge

PR #188, opened 2018-12-14
Allow SCope to be deployed on secure protocol (HTTPS)

mase5:
Update Apache config to deploy SCope on secure protocol (e.g.: HTTPS). Update related tutorials for AWS deployment. Update to version 1.3.5.


RafDeCoster:
Tested and works! thx!

Issue #192, opened 2019-02-05
Motif logos not shown if not from motif collection v8


mase5:
Motifs from motif collection versions different from `v8` are not shown in SCope.

KrisDavie:
Fixed in #211

PR #211, opened 2019-07-18
Allow motifs from aertslab v9

KrisDavie:
- Should be configured in loom in the future

Signed-off-by: KrisDavie <kristofer.davie@kuleuven.vib.be>


KrisDavie:
Fixes #192

PR #259, opened 2020-02-24
Maint/black

jhc-at-home:
Blacked server code


KrisDavie:
Seems to be good for me. How are we sitting on the line length? Do we want to open another branch to test shorter lines to get an idea on it, or are we happy as is?

jhc-at-home:
> Seems to be good for me. How are we sitting on the line length? Do we want to open another branch to test shorter lines to get an idea on it, or are we happy as is?

Hmmm. In general I'm fine leaving it at `120`. But what gives me a moment pause is a line in the `prettier` (docs)[https://prettier.io/docs/en/options.html#print-width]:
> Prettier, on the other hand, strives to fit the most code into every line. With the print width set to 120, prettier may produce overly compact, or otherwise undesirable code.

I'll run prettier  again in a moment (just finishing up CI before a PR) and see how bad it is.

jhc-at-home:
OK, `prettier` seems to do fine with a line length of `120` so I'm merging this.

PR #266, opened 2020-02-26
Bugfix/265 installation fails with 1.8.0

dweemx:


README.md
```diff
@@ -42,37 +41,9 @@ LOCAL_SCOPE_REPO="${HOME}/repos/SCope"
...
-```
-
-Create miniconda virtual environment with python (3.6 or higher) and nodejs (9 or higher) for SCope:
-
-```bash
-# Install python and nodejs with conda.
-conda create -n scope 'python==3.6.2' 'nodejs>=9'
-```
-
 #### Install SCope
```
jhc-at-home: Did you intend for this to be h4?

bin/install.js
```diff
@@ -4,20 +4,12 @@ console.log("Installing SCope Bind Server")
 shell.cd('opt/scopeserver/bindserver')
 shell.exec('npm install');
 shell.cd('../../../')
-var execSync = require('child_process').execSync;
```
jhc-at-home: Is this a dependency we can remove now? Great!!

opt/pyproject.toml
```diff
@@ -72,6 +68,13 @@ max-line-length = 120
 line-length = 120
 target-version = ['py38']
 
+[build-system]
+requires = [
+   "poetry>=0.12",
+   "setuptools==45.2.0"
+]
```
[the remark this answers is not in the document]
dweemx: Isn't better to fix package version since next version could potentially cause bugs (never know in advance) ?

bin/install.js
dweemx: No, unfortunately we still need to use it to run commands with `exec`

README.md
dweemx: I restructured README see 79a6582; better ?

opt/pyproject.toml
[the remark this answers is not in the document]
jhc-at-home: This won't change major version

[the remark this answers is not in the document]
dweemx: Ok

PR #273, opened 2020-03-06
Bugfix/272 error on getting regulon data

dweemx:


opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -704,14 +705,18 @@ def has_regulons_AUC(self) -> bool:
 
     def get_regulons_AUC(self, regulon_type: str = None):
         loom = self.loom_connection
-        if regulon_type == "motif":
+        if "MotifRegulonsAUC" in self.loom_connection.ca.keys() and regulon_type == "motif":
```
KrisDavie: Can you update the variables here from `L` to something i) Lower case ii) More descriptive?

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -704,14 +705,18 @@ def has_regulons_AUC(self) -> bool:
...
             loom.ca.MotifRegulonsAUC.dtype.names = list(map(lambda s: s.replace(" ", "_"), L))
             return loom.ca.MotifRegulonsAUC
-        if regulon_type == "track":
+        if "TrackRegulonsAUC" in self.loom_connection.ca.keys() and regulon_type == "track":
             L = loom.ca.TrackRegulonsAUC.dtype.names
             loom.ca.TrackRegulonsAUC.dtype.names = list(map(lambda s: s.replace(" ", "_"), L))
             return loom.ca.TrackRegulonsAUC
+        if "RegulonsAUC" in self.loom_connection.ca.keys():
+            L = loom.ca.RegulonsAUC.dtype.names
+            loom.ca.RegulonsAUC.dtype.names = list(map(lambda s: s.replace(" ", "_"), L))
```
KrisDavie: As we are moving away from `map`, something like `[s.replace(" ", "_") for s in L]` would be better here (and in the other statements)

KrisDavie:
LGTM. Merging with Develop to combine changes I am working on.

Will merge to master with both in same update.

PR #280, opened 2020-03-20
Add client side linting

jhc-at-home:
Turns off some linter rules to limit the number of required changes. These should be turned on later.


jhc-at-home:
CodeFactor seems to ignore the branch eslint config. Maybe taking from master? I don't know. Nevertheless, it's fine for now -- ignore the CodeFactor issues.

src/components/AppHeader.jsx
```diff
@@ -400,19 +401,19 @@ class AppHeader extends Component {
...
         );
         let d = pako.deflate(j, { to: 'string' });
         let b = encodeURIComponent(window.btoa(d).replace(/\//g, '$'));
-        if (this.state.permalinkUUID) {
-            var uuid = 'permalink__' + match.params.uuid;
-        } else {
-            var uuid = 'permalink';
-        }
+        const uuid =
+            'permalink' + (this.state.permalinkUUID ? match.params.uuid : '');
```
KrisDavie: This will break permalinks as is. 

```suggestion
            'permalink' + (this.state.permalinkUUID ? "__" + match.params.uuid : '');
```

jhc-at-home: Good catch! Thanks.

PR #282, opened 2020-03-25
Fix regression in b855e9d

jhc-at-home:
This fixes a regression introduced in b855e9d whereby an `image` variable is not in scope which crashes the application.


jhc-at-home:
> Tested locally heavy_check_mark

Great! Thanks! And good catch! :+1: 

I'll squash merge as soon as CI passes

PR #289, opened 2020-03-26
Correct the recursive run method

benzon21:
The run method was recursive, looks like an indented error


KrisDavie:
Hey @benzon21, nice catch! Thanks.

Could you run black over your branch so that it passes tests please, and then we can get this merged.

jhc-at-home:
Hi @benzon21, sorry for the bad experience you're having contributing to SCope, we're in the process of improving this experience. The issue is that black has not found our project configuration (`opt/pyproject.toml`). So we can merge your fix ASAP, could you please follow these steps:

1. Revert your previous commit. If you're using git in a terminal, this can be done with: `git revert HEAD`
2. Re-run black using our project configuration file. If you have `poetry` installed run: `poetry run black`. Otherwise, when you run `black` ensure you're in the `opt/` directory.

PR #291, opened 2020-04-02
Change deprecated use of react-file-downloader to js-file-downloader

jhc-at-home:


jhc-at-home:
Server CI failure is `safety`, this will be fixed by PR #292

PR #293, opened 2020-04-06
Change deprecated use of react-file-downloader to js-file-downloader

jhc-at-home:
`react-file-downloader` is no longer maintained.


src/components/common/Metadata.jsx
```diff
@@ -8,8 +8,11 @@ import {
     Loader,
     Dropdown
 } from 'semantic-ui-react';
-import { BackendAPI } from '../common/API';
 import ReactGA from 'react-ga';
+import { fileDownload } from 'js-file-download';
```
KrisDavie, line 12:
```suggestion
import fileDownload from 'js-file-download';
```
`fileDownload` is the default exported function and doesn't import and work correctly with the curly braces.

src/components/common/ViewerSidebar.jsx
```diff
@@ -12,21 +12,24 @@ import {
...
 import 'react-table/react-table.css';
-import FileDownloader from '../../js/http';
 import { instanceOf } from 'prop-types';
 import { withCookies, Cookies } from 'react-cookie';
 
-import CollaborativeAnnotation from './CollaborativeAnnotation';
-
 import { delimiter } from 'path';
 import { min } from 'moment';
+import { fileDownload } from 'js-file-download';
```
KrisDavie, line 26:
```suggestion
import fileDownload from 'js-file-download';
```
Same here as above

src/components/common/Metadata.jsx
jhc-at-home: Thanks!

PR #296, opened 2020-04-08
Search Refactor

### LINKED ISSUE #295: Allow searching of cell annotations
Make the categories for cell annotations searchable like the markers of a cluster are

KrisDavie:
Closes #295


opt/scopeserver/dataserver/modules/gserver/GServer.py
```diff
@@ -204,9 +204,24 @@ def get_features(self, loom: Loom, query: str):
         descriptions = {
             x: ""
             for x in collapsedResults.keys()
-            if x[1] not in ["region_gene_link", "regulon_target", "marker_gene", "cluster_annotation"]
+            if x[1]
+            not in ["region_gene_link", "regulon_target", "marker_gene", "cluster_annotation", "annotation_category"]
         }
 
+        def process_descriptions(
```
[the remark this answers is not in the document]
jhc-at-home: Here's how I might write it in 2 functions (ignoring obvious details)
```python
    def join_feature_descriptions():
        descriptions[(element, element_type)] = ", ".join(toProcess)
```

```python
    def add_result(element, element_type):
        pass # I don't understand what's going on here
```

opt/scopeserver/dataserver/modules/gserver/GServer.py
```diff
@@ -215,15 +230,18 @@ def get_features(self, loom: Loom, query: str):
                     synonyms.remove(r[0])
                 except ValueError:
                     pass
+                if r[1] == "annotation_category":
+                    for annotation in r[0]:
+                        collapsedResults, descriptions = process_descriptions(
+                            r, collapsedResults, descriptions, annotation, "annotation", "is a category of"
+                        )
+                    del collapsedResults[r]
```
jhc-at-home: :scream: please avoid this! Mutating the dictionary like this (`del`)

opt/scopeserver/dataserver/modules/gserver/GServer.py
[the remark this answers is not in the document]
jhc-at-home: It would also be nice if they had fewer parameters than the original function. It's ok to prepare parameters before giving the result to your function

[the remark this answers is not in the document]
jhc-at-home: `prepare_feature_descriptions` might also be a name worth considering

opt/scopeserver/dataserver/utils/search_space.py
```diff
@@ -168,6 +168,11 @@ def add_annotations(self) -> None:
         annotations = []
         for annotation in self.meta_data["annotations"]:
             annotations.append(annotation["name"])
+            for category in annotation["values"]:
+                if (category.casefold(), category, "annotation_category") in self.keys():
+                    self[(category.casefold(), category, "annotation_category")].append(annotation["name"])
+                else:
+                    self[(category.casefold(), category, "annotation_category")] = [annotation["name"]]
```
jhc-at-home, line 175:
```suggestion
                key = (category.casefold(), category, "annotation_category")
                if key in self:
                    self[key].append(annotation["name"])
                else:
                    self[key] = [annotation["name"]]
```

opt/scopeserver/dataserver/utils/search.py
```diff
@@ -0,0 +1,267 @@
...
+        for k, v in aggregated_matches.items():
+            if self.cross_species != "":
+                species_name = constant.SPECIES_MAP[self.cross_species]["short"]
+                identity_perc = self.cross_species_identity[k[0]]
+                descriptions[k] = f"Orthologue of {k[0]}, {identity_perc:.2f}% identity ({species_name} -> Drosophila)"
+            synonyms = v.copy()
+            try:
+                synonyms.remove(k[0])
+            except ValueError:
+                pass
```
jhc-at-home, line 194:
```suggestion
            synonyms = [synonym for synonym in v if synonym != k[0]]
```

opt/scopeserver/dataserver/utils/search.py
```diff
@@ -0,0 +1,267 @@
...
+
+        elif self.cross_species == "hsap":
+            for match in matches:
+                key = (match[1], match[2])
+                for droso_gene in self.dfh.hsap_to_dmel_mappings[match[0]]:
+                    if key not in aggregated_matches.keys():
+                        aggregated_matches[key] = [droso_gene[0]]
+                    else:
+                        aggregated_matches[key].append(droso_gene[0])
+                    self.cross_species_identity[match[0]] = droso_gene[1]
```
jhc-at-home, line 158: `match[0]` isn't being updated in this loop. This should be updated outside of this function anyway, it's not "aggregating matches"

opt/scopeserver/dataserver/utils/search.py
```diff
@@ -0,0 +1,267 @@
...
+        Collapse matches based on returned element.
+
+        match[0]: Match from search space
+        match[1]: Category of match
+        match[2]: Relationship
+        """
+
+        aggregated_matches: Dict[Tuple[str, str], List[str]] = OrderedDict()
+
+        if self.cross_species == "":
```
jhc-at-home, line 142: I would like this `if`/`else` block to be abstracted a bit more. Essentially it is grouping so `itertools.groupby()` will be useful.

opt/scopeserver/dataserver/utils/search.py
```diff
@@ -0,0 +1,267 @@
...
+        Search for matches in the search space.
+
+        Find keys in the search space where the search term matches (casefolded)
+        the searchable element. Return the match, the matches category and relationship.
+        """
+
+        if len(self.casefold_term) == 1:
+            matches = [x for x in self.search_space.keys() if self.casefold_term == x[0]]
+        else:
+            matches = [x for x in self.search_space.keys() if self.casefold_term in x[0]]
```
jhc-at-home, line 86:
```suggestion
        matches = [key for key in self.search_space if self.casefold_term in key[0]]
```

opt/scopeserver/dataserver/utils/search.py
jhc-at-home: Something like:
```python
def grouping_key(match: SearchMatch) -> Tuple[str, str]:
    return (match[1], match[2])

arguments = {
    "": (match for match in matches),
    "hsap": itertools.chain.from_iterable(
        (((droso_gene[0], *grouping_key(match)) for droso_gene
          in self.hsap_to_dmel_mappings[match[0]])
         for match in matches)
    ),
    "mmus": itertools.chain.from_iterable(
        (((droso_gene[0], *grouping_key(match)) for droso_gene
          in self.mmus_to_dmel_mappings[self.search_space[match]])
         for match in matches)
    )
}

groups = itertools.groupby(sorted(arguments[self.cross_species], key=grouping_key), key=grouping_key)
aggregated_matches = {key: [match[0] for match in group] for key, group in groups}
```
2 things for this to be viable:
1. why are the mappings being indexed differently?
2. I don't like how the data are being extracted from the mappings. Needs more thought.

opt/scopeserver/dataserver/utils/search.py
```diff
@@ -0,0 +1,267 @@
...
+
+        Some matches translate into something other than the term searched for. For example,
+        searching for a gene that is a marker of a cluster, will result in the searched gene
+        being translated into the name of the cluster that it is a marker of.
+        """
+
+        descriptions: Dict[Tuple[str, str], str] = {}
+
+        for k, v in aggregated_matches.items():
+            if self.cross_species != "":
```
jhc-at-home, line 186: This could be
```python
if self.cross_species is not None:`
```
If you decide that the Optional enum idea is the right way to go.

opt/scopeserver/dataserver/utils/search.py
```diff
@@ -0,0 +1,267 @@
...
+        features, feature_types = self.get_final_feature_and_type(aggregate_matches)
+        descriptions = self.create_feature_description(aggregate_matches, features)
+
+        final_res = {
+            "feature": [features[k] for k in aggregate_matches.keys()],
+            "featureType": [feature_types[k] for k in aggregate_matches.keys()],
+            "featureDescription": [descriptions[k] for k in aggregate_matches.keys()],
+        }
+
+        return final_res
```
jhc-at-home, line 267:
```suggestion
        final_res = {
            "feature": [features[k] for k in aggregate_matches],
            "featureType": [feature_types[k] for k in aggregate_matches],
            "featureDescription": [descriptions[k] for k in aggregate_matches],
        }

        return final_res
```

jhc-at-home:
In general, I'd like to see even more abstraction. Less magic number indexes. Fewer special case access. I'd prefer the interface changed: I'd like to avoid defining the Searcher class and I'd like you to be able to run the search with a single function call

KrisDavie:
A useful feature to add here would be a version in the search space pickle file. As this PR changes the search space, old pickles will no longer function correctly. Adding a version to the pickled object and checking for it on load would allow us to remove old pickles and regenerate them when needed.

opt/scopeserver/dataserver/utils/search.py
KrisDavie: The list comprehension is much slower than the copy + remove:
image

MaybeJustJames: Yeah that's to be expected. Is this performance critical? Pick what you like the most if you go with copy+remove consider this instead:
```python
from contextlib import suppress
#...
            with suppress(ValueError):
                synonyms.remove(k[0])
```
instead of
```python
            try:
                synonyms.remove(k[0])
            except ValueError:
                pass
```

KrisDavie: Nice, I'll stick with that

opt/scopeserver/dataserver/utils/search.py
```diff
@@ -0,0 +1,267 @@
...
+            + sorted(bad_match, key=sort_matches)
+        )
+
+        return final_matches
+
+    def aggregate_matches(self, matches: List[SearchMatch]) -> Dict[Tuple[str, str], List[str]]:
+        """
+        Collapse matches based on returned element.
+
+        match[0]: Match from search space
```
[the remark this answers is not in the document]
KrisDavie: Yep, this was a bit incorrect, cleaned this all up now.

opt/scopeserver/dataserver/utils/search.py
KrisDavie: If I split the identity extraction out of here, we'll have to re-loop through the matches, would you prefer I rename the function to `process_and_aggregate_matches` or something like that?

MaybeJustJames: In principle I would prefer you loop through the matches again. Is this likely to be expensive? If it's expensive this name change is fine.

opt/scopeserver/dataserver/utils/search.py
```diff
@@ -0,0 +1,323 @@
...
+        else:
+            match_results.append((match, (match[1], search_space[match], match[2])))
+
+    perfect_match: List[SearchMatch] = []
+    good_match: List[SearchMatch] = []
+    bad_match: List[SearchMatch] = []
+
+    for match_result in match_results:
+        # This contains the full match, its category and their relationship
+        match, result = match_result
```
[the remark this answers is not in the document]
MaybeJustJames: This loop is **really** trying to assign scores and you finish off by sorting based on the score. What do you think about a separate function, something like:
```python3
def match_result_cost(term: str, result: str) -> Optional[int]:
    if term.casefold() == result.casefold():
        return 0
    if result.casefold().startswith(term.casefold()) or \
       result.casefold().endswith(term.casefold()):
        return 1
    if term.casefold() in result.casefold():
        return 2

    return None

def sort_results(search_term: str, results: List[MatchResult]) -> List[SearchMatch]:
    costs = [(match_result_cost(search_term, match[1]), result) for match, result in results]
    sorted_costs = sorted([(cost, result) for cost, result in costs
                           if cost is not None], key=lambda cost: cost[0])
    return [result for _, result in sorted_costs]
```

Then, replace this loop and the remainder of this function with:
```python3
return sorted_results(search_term, match_results)
```

package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "SCope",
-    "version": "1.8.2",
+    "version": "1.8.3",
```
MaybeJustJames, line 3: Undo please

opt/scopeserver/dataserver/utils/search.py
```diff
@@ -0,0 +1,323 @@
...
+        search_space (SearchSpace): Search space from loom object
+
+    Returns:
+        List[SearchMatch]: A sorted list of the matches to the users search term
+    """
+
+    if len(casefold_term) == 1:
+        matches = [x for x in search_space if casefold_term == x[0]]
+    else:
+        matches = [x for x in search_space if casefold_term in x[0]]
```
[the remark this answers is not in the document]
KrisDavie: My code checks to see if a single letter was searched for, if so, we limit the search to only provide exact matches, otherwise searches take forever and return thousands of values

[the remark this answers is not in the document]
MaybeJustJames: Then the results need to be ranked (a la the `match_result_cost` function I defined in a later suggestion) and the top N "best" matches returned.

opt/scopeserver/dataserver/utils/search.py
KrisDavie: Not relevant now the cross species stuff is removed

opt/scopeserver/dataserver/utils/search.py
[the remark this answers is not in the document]
KrisDavie: This is very nice, it actually allows a much more fine-grained sorting also, without having to have a lot of extra lists

opt/scopeserver/dataserver/utils/search.py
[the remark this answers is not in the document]
KrisDavie: So this line is actually the most expensive operation and it increases with the number of results. For example, in one file, searching for `c` with just `in` (which matches 2898830 entries) takes ~4.5 seconds to find them and sort them, but if I use `==` for length 1 terms, this matches 2 results and takes ~6ms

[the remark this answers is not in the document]
ghuls: Not sure how helpful it would be for SCope, but pandas contains str functions that works on numpy arrays directly, which might be useful to find matches faster: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.Series.str.contains.html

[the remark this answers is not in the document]
MaybeJustJames: OK that's very interesting. `2_898_830` isn't a particularly big number. I'm curious if making `matches` a generator would change the performance at all:
```python3
matches = (x for x in search_space if casefold_term in x[0])
```

@ghuls idea to attack it with numpy is good in principle. A naive approach would be to transform the search space into an `ndarray` then filter that:
```python3
ss = np.array(search_space)
matches = ss[np.char.find(ss, casefold_term) != -1]
```

I measured it like so:
```python3
search_space = np.array([str(x) for x in range(3_000_000)])
timeit.timeit("search_space[np.char.find(search_space, '2') != -1]", number=100, globals={'search_space': search_space, 'np': np})
```
Each search takes ~1.48 seconds without sorting so the naive approach is no good either.

Alternatively using a `set()` might improve things:
```python3
search_space = set((str(x) for x in range(3_000_000)))
timeit.timeit("[x for x in search_space if '2' in x]", number=100, globals={'search_space': search_space})
```
Down to ~0.58 seconds without sorting. Not bad. Not 6ms.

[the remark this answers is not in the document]
KrisDavie: For now, I'm keeping the exception for 1 character searches, but switching to a generator did speed that step up also

PR #297, opened 2020-04-08
[WIP] Start a Flask app: install and begin restructure work

MaybeJustJames:
This is an initial implementation. You should be able to run the server and use it as you currently do now with:
`poetry run python scopeserver/app.py`


jhc-at-home:
At the moment the "`XServer`" is not being started. I need to fix this before a proper review.

PR #298, opened 2020-04-08
Speed up server CI

jhc-at-home:
Removes dependency on the pre-packaged github action for poetry.


jhc-at-home:
Please squash this branch before merging (if you're about to press the button).

PR #304, opened 2020-04-17
Feature/246 mvp app state tracking with redux

### LINKED ISSUE #246: Track application state using Redux
Install Redux and Redux Saga and set up a store according to the [getting started guide](https://redux.js.org/introduction/getting-started)

dweemx:
Closes #246


src/main.jsx
```diff
@@ -0,0 +1,58 @@
+import * as React from 'react';
```
jhc-at-home, line 1: Why is this not TypeScript?

package.json
```diff
@@ -39,10 +39,12 @@
     },
     "dependencies": {
         "@improbable-eng/grpc-web": "^0.12.0",
+        "@reduxjs/toolkit": "^1.3.4",
```
jhc-at-home, line 42: I'm not super happy about this but it might indeed make things a little easier. Can you also add `redux-saga` per the bug too please?

src/main.jsx
```diff
@@ -0,0 +1,58 @@
...
+    return {
+        isLoading
+    };
+};
+
+const actionCreators = {
+    setLoading
+};
+
+const ConnectedApp = connect(
```
jhc-at-home, line 33: This should not be here

src/main.jsx
dweemx: Because it was easier to not have TypeScript for the first trial. I was thinking of adding it after the first working trial

dweemx:
> Some changes needed. We weren't notified because you didn't add us as reviewers. I'm not adding reviewers for PRs that aren't ready... do I assume you're doing this too?

Ok my bad; yeah sounds good to me

src/main.jsx
dweemx: Where would you expect to see this instead?

MaybeJustJames: I'd like it to be broken into separate files and/or directories. There's a lot going on here.

src/main.jsx
MaybeJustJames: OK no worries. We'll make it `tsx` before merging then.

package.json
MaybeJustJames: I'm actually quite against `toolkit` now. I don't think we need what it does. Please remove (see https://react-redux.js.org/introduction/basic-tutorial) for a basic redux guide

src/main.jsx
dweemx: Done (seems that just the types were missing)

opt/poetry.lock
```diff
@@ -867,7 +867,7 @@ category = "main"
 description = "YAML parser and emitter for Python"
 name = "pyyaml"
 optional = false
-python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
+python-versions = "*"
```
MaybeJustJames, line 870: This isn't right. Actually it should probably be
```suggestion
python-versions = ">=3.7"
```

src/components/App.jsx
```diff
@@ -33,6 +34,8 @@ import About from './pages/About';
 import Alert from 'react-popup';
 import 'react-popup/style.css';
 
+import { setLoading as reduxAction_setLoading } from '../redux/actions';
```
MaybeJustJames, line 37: This isn't a great name, and `reduxAction_setLoading` is even worse. How about `setAppLoading`?

src/components/App.jsx
```diff
@@ -87,14 +89,15 @@ class App extends Component {
...
         const {
-            loading,
             metadata,
             error,
             loaded,
             isSidebarVisible,
             sessionsLimitReached,
         } = this.state;
 
+        const { isLoading } = this.props;
```
MaybeJustJames, line 99: I expected the `Component` to be parameterised by a custom prop type definition?

opt/poetry.lock
MaybeJustJames: In fact, this PR shouldn't be touching `opt/` please revert.

src/components/App.jsx
dweemx: Yeah probably better

src/redux/reducers/main.ts
```diff
@@ -1,22 +1,23 @@
-import { SET_LOADING } from '../actionTypes';
+import { SET_APP_LOADING } from '../actionTypes';
 import { SetLoadingAction } from '../types';
 
 type Loading = {
```
MaybeJustJames, line 4: Maybe this should be `AppState`. You've even already defined an interface `AppLoading` that you didn't import from `../types`.
So, change the interface name to `AppState` (or make it an alias, I don't mind) or something more generic than `...Loading`.

opt/poetry.lock
```diff
@@ -229,24 +229,6 @@ toml = "*"
 [package.extras]
```
MaybeJustJames, line 229: This file shouldn't appear in the diff at all

src/components/App.jsx
```diff
@@ -1,4 +1,4 @@
-import { instanceOf } from 'prop-types';
+import { PropTypes, instanceOf } from 'prop-types';
```
MaybeJustJames, line 1: This isn't actually what I meant, maybe it's ok...
I meant you define a properties alias/interface like [here](https://github.com/aertslab/SCope/pull/363/files#diff-cc751c9a9e01ff5fb6d73e2e6128074aR28) and parameterise `Component` with that type.

I'm not actually sure if `PropTypes` does validation. Do you know?

MaybeJustJames: Oh I'm an idiot. Sorry I didn't realise this was still JS, of course! In that case this is fine.

PR #306, opened 2020-05-05
Bugfix/305 large dataset rendering

### LINKED ISSUE #305: Large datasets do not display all cells
Datasets > 200k cells do not render all points due to the PartcileContainer being limited in size at this line:
https://github.com/aertslab/SCope/blob/0468e2f6c9ab7791ad1ebbd27ddaeb433b9251bc/src/components/common/Viewer.jsx#L51

PIXI has an option to dynamically adjust the size of the container, however this destroys performance. We should adjust the container size when we receive the points from the frontend, this is the first time we know how may cells there will be.

KrisDavie:
Fixes #305 
Also fixes a longstanding bug where points do not display when someone directly visits a tabbed view

### DIFF
--- a/src/components/common/Viewer.jsx
+++ b/src/components/common/Viewer.jsx
@@ -48,7 +48,7 @@ export default class Viewer extends Component {
         this.w = parseInt(this.props.width);
         this.h = parseInt(this.props.height);
         // Increase the maxSize if displaying more than 1500 (default) objects
-        this.maxn = 200000;
+        this.maxn = 2000;
         this.texture = PIXI.Texture.fromImage('src/images/dot.png');
         this.settingsListener = (settings, customScale) => {
             if (this.props.settings) {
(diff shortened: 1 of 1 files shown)


src/components/common/API.jsx
```diff
@@ -292,13 +292,26 @@ class API {
...
+        return new Promise((resolve) => {
+            if (
+                typeof this.loomFiles !== 'undefined' &&
+                this.activeLooms[0] in this.loomFiles
+            ) {
+                resolve(
+                    this.loomFiles[this.activeLooms[0]].cellMetaData.embeddings
+                );
+            } else {
+                setTimeout(this.getActiveLoomMetadataEmbeddings, 250);
```
jhc-at-home, line 304: My only (minor - for now) concern is that this will never end.

KrisDavie:
Rewrote this branch to only include the fix for the rendering bug. Opening another, detailed bug for the tabbed view thing

src/components/common/API.jsx
KrisDavie: See comment below, this was flawed anyway and should be addressed in its own PR

PR #308, opened 2020-05-07
Reorganise and abstract SCope server config

MaybeJustJames:
* Creates an "outer layer" config that the "inner layer" business code can depend on.


MaybeJustJames:
I don't like the `scope` package name. But I can't think of a better one yet

opt/scopeserver/__init__.py
```diff
@@ -69,45 +70,20 @@ def run(self) -> None:
...
- \____  $$| $$      | $$  \ $$| $$  \ $$| $$$$$$$$       \____  $$| $$$$$$$$| $$  \__/ \  $$/$$/| $$$$$$$$| $$  \__/
- /$$  \ $$| $$    $$| $$  | $$| $$  | $$| $$_____/       /$$  \ $$| $$_____/| $$        \  $$$/ | $$_____/| $$
-|  $$$$$$/|  $$$$$$/|  $$$$$$/| $$$$$$$/|  $$$$$$$      |  $$$$$$/|  $$$$$$$| $$         \  $/  |  $$$$$$$| $$
- \______/  \______/  \______/ | $$____/  \_______/       \______/  \_______/|__/          \_/    \_______/|__/
-                              | $$
-                              | $$
-                              |__/
-        """
-    )
+    with open(Path("data") / Path("motd.txt")) as motd:
```
KrisDavie, line 73:
```suggestion
    with open(Path(__file__).parents[1] / Path("data") / Path("motd.txt")) as motd:
```

This only works when running from opt, which is a bit annoying. Resolving the path relative to the init script will fix this.

opt/scopeserver/__init__.py
```diff
@@ -1,19 +1,20 @@
...
 
+from scopeserver.dataserver.modules.gserver import GServer as gs
+from scopeserver.dataserver.modules.pserver import PServer as ps
+from scopeserver.bindserver import XServer as xs
+from scopeserver.dataserver.utils import sys_utils as su
+from scopeserver.scope.config import from_file
+
 LOG_FMT = "%(asctime)s - %(name)s - %(levelname)s - %(message)s"
 logging.basicConfig(format=LOG_FMT, level=logging.INFO)
 logger = logging.getLogger(__name__)
```
KrisDavie, line 20: `logger` should be changes to `LOGGER` to keep consistency with the other changes here.

opt/scopeserver/__init__.py
```diff
@@ -69,45 +70,20 @@ def run(self) -> None:
...
-            except json.JSONDecodeError:
-                config = {}
-                logger.error("Config file is not proper json and will not be used")
-    else:
-        config = {}
-
-    dhs = config.get("dataHashSecret")
-
-    if not dhs or not dhs.strip():
+    if config.get("dataHashSecret") is None:
```
KrisDavie, line 80: This change allows an empty string to be valid as a secret (which negates its use), the `dhs.strip()` check prevents this.

MaybeJustJames: I've added validation in the `config` module for this

opt/scopeserver/__init__.py
MaybeJustJames: Done. Thanks

opt/scopeserver/__init__.py
MaybeJustJames: I've added a config item called `"data"` to find this file (defaults to "data/")

MaybeJustJames:
@KrisDavie this is ready for you to have another look at it

opt/scopeserver/__init__.py
KrisDavie: For the `"data"` config option, you should add this to the highest level `config.json`. However, this will still fail to run on a new install which hasn't been configured or when running outside of the `opt` dir, it is such an irrelevant bit to fail on. 

Can we add try/except for the file actually existing? Either falling back to not showing the MOTD, or resolving the relative path like in the suggestion above?

MaybeJustJames: Ah good point. I've added an `is_file()` existence check with a fallback to, "Welcome to SCope." if it doesn't exist.

KrisDavie: Thanks. Last things is to add something to document the entry in `config.json` in the root of the repo. A default is good or also an empty key would work.

MaybeJustJames: Done. Thanks

PR #327, opened 2020-05-14
Enable more clusters to be visualised

KrisDavie:
Looping through the list should still provide visual distinction, rather than arbitrarily increasing the number of colours

### DIFF
--- a/opt/scopeserver/dataserver/utils/cell_color_by_features.py
+++ b/opt/scopeserver/dataserver/utils/cell_color_by_features.py
@@ -172,19 +172,19 @@ class CellColorByFeatures:
                 clusteringID = str(clustering["id"])
                 if request.feature[n] == "All Clusters":
                     numClusters = max(self.loom.get_clustering_by_id(clusteringID))
-                    if numClusters <= 245:
-                        self.legend = set()
-                        clustering_meta = self.loom.get_meta_data_clustering_by_id(int(clusteringID), secret=secret)
-                        cluster_dict = {int(x["id"]): x["description"] for x in clustering_meta["clusters"]}
-                        for i in self.loom.get_clustering_by_id(clusteringID):
-                            self.hex_vec.append(constant.BIG_COLOR_LIST[i])
-                            self.legend.add((cluster_dict[i], constant.BIG_COLOR_LIST[i]))
-                        self.legend = s_pb2.ColorLegend(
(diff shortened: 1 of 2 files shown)


opt/scopeserver/dataserver/utils/cell_color_by_features.py
```diff
@@ -172,19 +172,19 @@ def setClusteringFeature(self, request, feature, n, secret=None):
...
-                        interval = int(RGB_COLORS / numClusters)
-                        self.hex_vec = [hex(i)[2:].zfill(6) for i in range(0, numClusters, interval)]
+                    self.legend = set()
+                    clustering_meta = self.loom.get_meta_data_clustering_by_id(int(clusteringID), secret=secret)
+                    cluster_dict = {int(x["id"]): x["description"] for x in clustering_meta["clusters"]}
+                    for i in self.loom.get_clustering_by_id(clusteringID):
+                        if i >= len(constant.BIG_COLOR_LIST):
+                            colour = constant.BIG_COLOR_LIST[i % len(constant.BIG_COLOR_LIST)]
+                        else:
+                            colour = constant.BIG_COLOR_LIST[i]
```
MaybeJustJames, line 182:
```suggestion
                        colour = constant.BIG_COLOR_LIST[i % len(constant.BIG_COLOR_LIST)]
```

opt/scopeserver/dataserver/utils/cell_color_by_features.py
```diff
@@ -172,19 +172,19 @@ def setClusteringFeature(self, request, feature, n, secret=None):
...
+                        self.hex_vec.append(colour)
+                        self.legend.add((cluster_dict[i], colour))
+                    self.legend = s_pb2.ColorLegend(
+                        values=[x[0] for x in self.legend], colors=[x[1] for x in self.legend]
```
MaybeJustJames, line 186: How large could `self.legend` get? If more than a hundred values or so then it's probably a good place to use a generator expression here unless protobuf is explicitly requesting lists:
```suggestion
                        values=(x[0] for x in self.legend), colors=(x[1] for x in self.legend)
```
That being said, we go to a lot of trouble to package `values` and `colors` up together just to unpack them when we want to use them. Are you using a `set()` because `cluster_dict` contains duplicate values? In that case, put `cluster_dict` values in a **local** set (don't reuse the class scope `legend`). Then slice into `self.hex_vec` to the `len()` of the values set.

PR #329, opened 2020-05-15
Fix wrong data type and uninitialised variable

KrisDavie:


MaybeJustJames:
Feel free to merge if you're happy with it.

KrisDavie:
Yeah, I have no idea why mypy didn't catch the second issue. The variable was my fault,

PR #330, opened 2020-05-18
Bugfix/312 next previous buttons

### LINKED ISSUE #312: Next/Previous buttons skip between clusterings
Using the next or previous buttons in the annotation section seems to skip back to the lowest index clustering, likely a hard coded index?

KrisDavie:
Fixes #312

### DIFF
--- a/opt/scopeserver/dataserver/modules/gserver/GServer.py
+++ b/opt/scopeserver/dataserver/modules/gserver/GServer.py
@@ -400,10 +400,18 @@ class SCope(s_pb2_grpc.MainServicer):
         clustering_meta = loom.get_meta_data_clustering_by_id(
             request.clusteringID, secret=self.config["dataHashSecret"]
         )
+        max_cluster = max([int(x["id"]) for x in clustering_meta["clusters"]])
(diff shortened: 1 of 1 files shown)


opt/scopeserver/dataserver/modules/gserver/GServer.py
```diff
@@ -412,10 +423,16 @@ def getNextCluster(self, request, context):
             )
 
         f = self.get_features(loom=loom, query=cluster_metadata["description"])
+
+        for n, featureType in enumerate(f["featureType"]):
+            if featureType == f"Clustering: {clustering_meta['name']}":
+                clustering_index = n
+                break
```
MaybeJustJames, line 430: You could consider replacing this with a function, something like, `get_feature_index()`

KrisDavie: I'm going to leave this as is at the moment, because I would like to overhaul this in the future. Relying on the 'name' field is bad because there is no guarantee for uniqueness. We strip a lot of information when sending responses to the fronted, like the ID of the clustering, which would be very useful in these circumstances if we kept it. This would eliminate the need for this code or the suggested function. Will add an issue about this.

PR #341, opened 2020-05-25
Fixes #340

### LINKED ISSUE #340: Cell embeddings are not visible with a scaling factor less than 1

dweemx:


src/components/common/Viewer.jsx
```diff
@@ -946,6 +946,12 @@ export default class Viewer extends Component {
             this.renderer.height /
             (d3.max(this.state.coord.y) - d3.min(this.state.coord.y));
         this.scalingFactor = Math.floor(d3.min([min, max])) - 1;
+        if (this.scalingFactor <= 0) {
+            console.warn(
+                '[WARNING] Scaling factor is zero or negative! The factor will be set to 1.'
+            );
+            this.scalingFactor = 1;
+        }
```
MaybeJustJames, line 954: It looks like you're trying to `clamp` values within a range? The minimum value in the range should be `1`?

```javascript
import * as R from 'ramda';`
this.scalingFactor = R.clamp(1, max_of_range, Math.floor(d3.min([min, max])) - 1);
```

MaybeJustJames: Is the log really useful? Under what conditions does the scalingFactor become less than 0?

dweemx: Maybe not, I can remove it
I don't think we ensure that min max won't be zero. In that case, scaling factor would be 0.
Shouldl I overwrite with your `Ramda` solution ?

MaybeJustJames: Looking at where `min` and `max` is defined makes me more confused. They _look_ like divisions by zero. Assuming they're not, is the minimum scaling factor 0 + epsilon? I'd be inclined to pick a minimum (say 0.001) and write it with:
```javascript
const MIN_SCALING_FACTOR = 0.001;
this.scalingFactor = R.clamp(MIN_SCALING_FACTOR, Infinity, Math.floor(Math.min(min, max)) - 1);
```

MaybeJustJames: Also they're not really `min` or `max`. They're horizontal and vertical scaling factors. Does it make sense to take the `max` of `this.state.coord.x`?

package.json
```diff
@@ -113,7 +113,6 @@
         "eslint-config-prettier": "^6.11.0",
         "eslint-plugin-prettier": "^3.1.3",
         "eslint-plugin-react": "^7.19.0",
-        "file-system": "^2.2.2",
```
MaybeJustJames, line 116: :smile: excellent!

PR #346, opened 2020-06-05
Catch error when gene not present from regulon

### LINKED ISSUE #345: Regulon data does not display when regulon gene is not present
This is mainly an issue for scATAC data, where regions are _linked_ to genes which are not actually _present_ in the loom file. 

This hangs up the backend and doesn't present metadata, and blocks showing AUC values and thresholded cells.

KrisDavie:
Fixes #345


KrisDavie:
Great, thanks!

PR #350, opened 2020-06-12
Feature/349 ss not rebuilding when files overwritten on back-end

### LINKED ISSUE #349: Search space not rebuilding when files are overwritten directly from back-end

dweemx:
Fixes #349

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,4 @@ assets
 release/*
 release/logs
 debug
+tests/
--- a/opt/pyproject.toml
+++ b/opt/pyproject.toml
@@ -51,6 +51,7 @@ requests = "^2.22.0"
 fastapi = "^0.54.1"
 hypercorn = "^0.9.5"
 methodtools = "^0.1.2"
+imohash = "^1.0.4"
 
 [tool.poetry.dev-dependencies]
 pytest = "*"
--- a/opt/scopeserver/dataserver/utils/loom_file_handler.py
+++ b/opt/scopeserver/dataserver/utils/loom_file_handler.py
@@ -120,20 +120,29 @@ class LoomFileHandler:
         # To check if the given file path is given specified url!
         file_hash = LoomFileHandler.get_file_hash(abs_loom_file_path)
         if abs_loom_file_path in self.active_looms:
-            logger.debug("Should be preloaded")
-            try:
-                logger.debug(
-                    f"Current mode: {self.active_looms[abs_loom_file_path].get_connection().mode}, wanted mode {mode}"
-                )
-
-                if self.active_looms[abs_loom_file_path].get_connection().mode == mode:
-                    loom = self.active_looms[abs_loom_file_path]
+            current_loom_file_hash = self.active_looms[abs_loom_file_path].get_file_hash()
+            if current_loom_file_hash != file_hash:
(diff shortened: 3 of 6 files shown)


opt/pyproject.toml
```diff
@@ -51,6 +51,7 @@ requests = "^2.22.0"
 fastapi = "^0.54.1"
 hypercorn = "^0.9.5"
 methodtools = "^0.1.2"
+imohash = "^1.0.4"
```
MaybeJustJames, line 54: I think this is a _very_ unwise change. Why not use `hashlib`? It's builtin

.gitignore
```diff
@@ -49,3 +49,4 @@ assets
 release/*
 release/logs
 debug
+tests/
```
MaybeJustJames, line 52: Why? These should be in git.

opt/scopeserver/dataserver/utils/loom_file_handler.py
```diff
@@ -1,5 +1,6 @@
 import os
 import hashlib
+import imohash
```
MaybeJustJames, line 3:
```suggestion
```

opt/pyproject.toml
MaybeJustJames: Also from `imohash` docs:
# Misuses
Because imohash only reads a small portion of a file's data, it is not suitable for:
* file verification or integrity monitoring

opt/scopeserver/dataserver/utils/loom_file_handler.py
```diff
@@ -34,12 +35,16 @@ def load_loom_file(self, partial_md5_hash: str, file_path: str, abs_file_path: s
...
-            partial_md5_hash=partial_md5_hash,
+            file_hash=file_hash,
             file_path=file_path,
             abs_file_path=abs_file_path,
             loom_connection=loom_connection,
         )
 
+    @staticmethod
+    def get_file_hash(file_path: str):
+        return imohash.hashfile(file_path, hexdigest=True)
```
MaybeJustJames, line 43: I don't really understand what this was trying to achieve in the old version either @KrisDavie ?
I realise timestamps etc might change and you don't want to change the hash if that changes... what is the high-level goal you're trying to achieve? *And* what's the bug in the other project that this is supposed to fix?
```suggestion
        hash = hashlib.md5()
        with open(file_path, "rb") as f:
            buffer = f.read(1024) # or something...
            while len(buffer) > 0:
                hash.update(buffer)
                buffer = f.read(1024)
        return hash.hexdigest()
```

MaybeJustJames (changes requested):
I can't accept this as-is. Please see comments.

.gitignore
dweemx: I have a local folder (for fast testing) with currently 1 ipynb that's not relevant to add to the github

MaybeJustJames: OK, the name's too generic though. Maybe `.local-tests/` or `.private-tests/`?

MaybeJustJames: Actually, could you add it to our tests? This is exactly what tests are for

MaybeJustJames:
Can this PR be closed?

dweemx:
> Can this PR be closed?

I guess since we are not going in this direction

PR #360, opened 2020-06-19
Remove reliance on file hashes

KrisDavie:
- Remove all hash code
- Add thread locking
- Add search space update code
- Various code cleanups

This is not a backwards breaking change, but could change workflows that people are used to. Loom files can now no longer be updated directly via the backend (adding a new file still works). Deleting looms will also cause problems if the search space is not also removed.

Old search spaces will not be automatically deleted, but won't be used either

### DIFF
--- a/opt/scopeserver/dataserver/utils/loom.py
+++ b/opt/scopeserver/dataserver/utils/loom.py
@@ -45,7 +45,7 @@ class Loom:
         self.ss_pickle_name = self.abs_file_path + ".ss_pkl"
         self.load_ss(self.ss_pickle_name)
 
-    def load_ss(self, ss_pickle_name: str):
+    def load_ss(self, ss_pickle_name: str) -> None:
         try:
             with open(ss_pickle_name, "rb") as fh:
                 logger.debug(f"Loading prebuilt SS for {self.file_path} from {ss_pickle_name}")
@@ -55,7 +55,9 @@ class Loom:
                 try:
                     assert self.ss.search_space_version == ss.CURRENT_SS_VERISON
                 except AssertionError:
-                    logger.error(f"Cached search space version {self.ss.search_space_version} is not {ss.CURRENT_SS_VERISON}. Rebuilding search space...")
+                    logger.error(
+                        f"Cached search space version {self.ss.search_space_version} is not {ss.CURRENT_SS_VERISON}. Rebuilding search space..."
+                    )
                     self.build_ss()
                 except AttributeError:
                     logger.error(f"Search space has no version key and is likely legacy. Rebuilding search space...")
@@ -69,7 +71,7 @@ class Loom:
         logger.debug(f"Built Search Space for {self.file_path}")
         self.write_ss()
 
-    def write_ss(self):
+    def write_ss(self) -> None:
         self.ss.loom = None  # Remove loom connection to enable pickling
         self.ss.dfh = None
         logger.debug(f"Built all Search Spaces for {self.file_path}")
@@ -77,7 +79,7 @@ class Loom:
             logger.debug(f"Writing prebuilt SS for {self.file_path} to {self.ss_pickle_name}")
             pickle.dump(self.ss, fh)
 
-    def update_ss(self, update: str):
+    def update_ss(self, update: str) -> None:
         self.ss.loom = self
         self.ss.meta_data = self.get_meta_data()
         if update == "clusterings":
--- a/opt/scopeserver/dataserver/utils/loom.py
+++ b/opt/scopeserver/dataserver/utils/loom.py
@@ -11,6 +11,8 @@ import pickle
 from copy import deepcopy
 from pathlib import Path
 from typing import Tuple, Dict, Any, Union, List, Set
+from loompy.loompy import LoomConnection
+from scopeserver.dataserver.utils.loom_file_handler import LoomFileHandler
 from typing_extensions import TypedDict
 from google.protobuf.internal.containers import RepeatedScalarFieldContainer
 
@@ -33,7 +35,9 @@ Clustering = TypedDict(
 
 
 class Loom:
-    def __init__(self, file_path, abs_file_path, loom_connection, loom_file_handler):
+    def __init__(
+        self, file_path: Path, abs_file_path: Path, loom_connection: LoomConnection, loom_file_handler: LoomFileHandler
+    ):
         self.lfh = loom_file_handler
         self.file_path = file_path
         self.abs_file_path = abs_file_path
--- a/opt/scopeserver/dataserver/utils/loom.py
+++ b/opt/scopeserver/dataserver/utils/loom.py
@@ -8,6 +8,7 @@ import time
 import hashlib
 import os
 import pickle
+from pathlib import Path
 from typing import Tuple, Dict, Any, Union, List, Set
 from typing_extensions import TypedDict
 from google.protobuf.internal.containers import RepeatedScalarFieldContainer
@@ -42,10 +43,10 @@ class Loom:
         # Metrics
         self.nUMI = None
         self.species, self.gene_mappings = self.infer_species()
-        self.ss_pickle_name = self.abs_file_path + ".ss_pkl"
+        self.ss_pickle_name = self.abs_file_path.with_suffix(".ss_pkl")
         self.load_ss(self.ss_pickle_name)
 
-    def load_ss(self, ss_pickle_name: str) -> None:
+    def load_ss(self, ss_pickle_name: Path) -> None:
         try:
             with open(ss_pickle_name, "rb") as fh:
                 logger.debug(f"Loading prebuilt SS for {self.file_path} from {ss_pickle_name}")
@@ -91,10 +92,10 @@ class Loom:
     def get_connection(self):
         return self.loom_connection
 
-    def get_file_path(self) -> str:
+    def get_file_path(self) -> Path:
         return self.file_path
 
-    def get_abs_file_path(self) -> str:
+    def get_abs_file_path(self) -> Path:
         return self.abs_file_path
 
     def get_global_attribute_by_name(self, name):
--- a/opt/scopeserver/dataserver/utils/loom.py
+++ b/opt/scopeserver/dataserver/utils/loom.py
@@ -8,6 +8,7 @@ import time
 import hashlib
 import os
 import pickle
+from copy import deepcopy
 from pathlib import Path
 from typing import Tuple, Dict, Any, Union, List, Set
 from typing_extensions import TypedDict
@@ -75,10 +76,11 @@ class Loom:
     def write_ss(self) -> None:
         self.ss.loom = None  # Remove loom connection to enable pickling
         self.ss.dfh = None
+        cur_ss = deepcopy(self.ss)
         logger.debug(f"Built all Search Spaces for {self.file_path}")
         with open(self.ss_pickle_name, "wb") as fh:
             logger.debug(f"Writing prebuilt SS for {self.file_path} to {self.ss_pickle_name}")
-            pickle.dump(self.ss, fh)
+            pickle.dump(cur_ss, fh)
 
     def update_ss(self, update: str) -> None:
         self.ss.loom = self
--- a/opt/scopeserver/dataserver/utils/loom.py
+++ b/opt/scopeserver/dataserver/utils/loom.py
@@ -428,11 +428,10 @@ class Loom:
 
         self.update_ss(update="clusterings")
 
-        try:
-            _ = self.get_clustering_by_id(new_clustering_meta["id"])
+        if new_clustering_meta["id"] in self.loom_connection.ca.Clusterings.dtype.names:
             logger.debug("Success")
             return (True, "Success")
-        except IndexError:
+        else:
             logger.debug("Failure")
             return (False, "Updated clusterings do not exist in file. Write error.")
 
--- a/opt/scopeserver/dataserver/utils/loom.py
+++ b/opt/scopeserver/dataserver/utils/loom.py
@@ -6,9 +6,6 @@ from methodtools import lru_cache
 import pandas as pd
 import time
 import hashlib
-import os
-import pickle
-from copy import deepcopy
 from pathlib import Path
 from typing import Tuple, Dict, Any, Union, List, Set
 from loompy.loompy import LoomConnection
@@ -48,51 +45,7 @@ class Loom:
         self.nUMI = None
         self.species, self.gene_mappings = self.infer_species()
         self.ss_pickle_name = self.abs_file_path.with_suffix(".ss_pkl")
-        self.load_ss(self.ss_pickle_name)
-
-    def load_ss(self, ss_pickle_name: Path) -> None:
-        try:
-            with open(ss_pickle_name, "rb") as fh:
-                logger.debug(f"Loading prebuilt SS for {self.file_path} from {ss_pickle_name}")
-                self.ss = pickle.load(fh)
-                self.ss.loom = self
-                self.ss.dfh = dfh.DataFileHandler()
-                try:
-                    assert self.ss.search_space_version == ss.CURRENT_SS_VERISON
-                except AssertionError:
-                    logger.error(
-                        f"Cached search space version {self.ss.search_space_version} is not {ss.CURRENT_SS_VERISON}. Rebuilding search space..."
-                    )
-                    self.build_ss()
-                except AttributeError:
-                    logger.error(f"Search space has no version key and is likely legacy. Rebuilding search space...")
-                    self.build_ss()
-        except (EOFError, FileNotFoundError):
-            self.build_ss()
-
-    def build_ss(self):
-        logger.debug(f"Building Search Spaces for {self.file_path}")
-        self.ss = ss.SearchSpace(loom=self).build()
-        logger.debug(f"Built Search Space for {self.file_path}")
-        self.write_ss()
-
-    def write_ss(self) -> None:
-        self.ss.loom = None  # Remove loom connection to enable pickling
-        self.ss.dfh = None
-        cur_ss = deepcopy(self.ss)
-        logger.debug(f"Built all Search Spaces for {self.file_path}")
-        with open(self.ss_pickle_name, "wb") as fh:
-            logger.debug(f"Writing prebuilt SS for {self.file_path} to {self.ss_pickle_name}")
-            pickle.dump(cur_ss, fh)
-
-    def update_ss(self, update: str) -> None:
-        self.ss.loom = self
-        self.ss.meta_data = self.get_meta_data()
-        if update == "clusterings":
-            self.ss.add_clusterings()
-        elif update == "cluster_annotations":
-            self.ss.add_cluster_annotations()
-        self.write_ss()
+        self.ss = ss.load_ss(self)
 
     def get_connection(self):
         return self.loom_connection
@@ -164,7 +117,7 @@ class Loom:
         metaJson["clusterings"][clustering_n]["clusters"][cluster_n]["description"] = new_annotation_name
         self.update_metadata(metaJson)
 
-        self.update_ss(update="clusterings")
+        ss.update_ss(self, update="clusterings")
 
         if (
             self.get_meta_data()["clusterings"][clustering_n]["clusters"][cluster_n]["description"]
@@ -246,7 +199,7 @@ class Loom:
 
         self.update_metadata(metaJson)
 
-        self.update_ss(update="cluster_annotations")
+        ss.update_ss(self, update="cluster_annotations")
 
         if (
             cell_type_annotation
@@ -431,7 +384,7 @@ class Loom:
         loom.attrs["MetaData"] = json.dumps(metaJson)
         self.loom_connection = self.lfh.change_loom_mode(self.file_path, mode="r")
 
-        self.update_ss(update="clusterings")
+        ss.update_ss(self, update="clusterings")
 
         if new_clustering_meta["id"] in self.loom_connection.ca.Clusterings.dtype.names:
             logger.debug("Success")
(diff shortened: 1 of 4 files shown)


MaybeJustJames:
This is subtle in many many ways. I can't review today. I'll have a look tomorrow

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -397,19 +416,16 @@ def add_user_clustering(self, request) -> Tuple[bool, str]:
...
-            self.file_path, mode="r", partial_md5_hash=self.partial_md5_hash
-        )
-        loom = self.loom_connection
+        self.loom_connection = self.lfh.change_loom_mode(self.file_path, mode="r")
+
+        self.update_ss(update="clusterings")
 
         try:
-            new_clustering_data = self.get_clustering_by_id(new_clustering_meta["id"])
+            _ = self.get_clustering_by_id(new_clustering_meta["id"])
```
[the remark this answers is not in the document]
KrisDavie: The function attempts to get the newly added id. We've had issues in the past where despite everything looking OK, the file didn't actually accept changes, so this is a sanity check to make sure the new clustering is actually present and readable.

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -397,19 +416,16 @@ def add_user_clustering(self, request) -> Tuple[bool, str]:
...
         loom = self.loom_connection
         loom.ca.Clusterings = new_clusterings
         loom.attrs["MetaData"] = json.dumps(metaJson)
-        self.loom_connection = self.lfh.change_loom_mode(
-            self.file_path, mode="r", partial_md5_hash=self.partial_md5_hash
-        )
-        loom = self.loom_connection
+        self.loom_connection = self.lfh.change_loom_mode(self.file_path, mode="r")
+
+        self.update_ss(update="clusterings")
```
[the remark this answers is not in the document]
KrisDavie: No, the update now only adds the new entries to the search space rather than rebuilding the entire thing (see https://github.com/aertslab/SCope/pull/360/files#diff-4cb5f7211d42fdedde8dab51db72b61cR79-R86)

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -43,29 +42,48 @@ def __init__(self, partial_md5_hash, file_path, abs_file_path, loom_connection,
...
             with open(self.ss_pickle_name, "rb") as fh:
-                logger.debug(f"Loading prebuilt SS for {file_path} from {self.ss_pickle_name}")
+                logger.debug(f"Loading prebuilt SS for {self.file_path} from {self.ss_pickle_name}")
                 self.ss = pickle.load(fh)
-
+            logger.debug(f"Building hsap Search Spaces for {self.file_path}")
+            self.hsap_ss = ss.SearchSpace(loom=self, cross_species="hsap").build()
+            logger.debug(f"Building mmus Search Spaces for {self.file_path}")
+            self.mmus_ss = ss.SearchSpace(loom=self, cross_species="mmus").build()
         except (EOFError, FileNotFoundError):
```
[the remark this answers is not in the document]
KrisDavie: Which bit? Writing the SS to disc?

opt/scopeserver/dataserver/utils/loom_file_handler.py
```diff
@@ -1,6 +1,10 @@
 import os
-import hashlib
 import loompy as lp
+from loompy import LoomConnection
+import threading
+import collections as c
```
dweemx, line 5:
```suggestion
from collections import defaultdict
```

opt/scopeserver/dataserver/utils/loom_file_handler.py
```diff
@@ -12,65 +16,38 @@
 class LoomFileHandler:
     def __init__(self):
         self.active_looms = {}
-        self.loom_dir = dfh.DataFileHandler.get_data_dir_path_by_file_type(file_type="Loom")
+        self.file_locks = c.defaultdict(lambda: threading.Lock())
```
dweemx, line 19:
```suggestion
        self.file_locks = defaultdict(lambda: threading.Lock())
```

opt/scopeserver/dataserver/utils/loom_file_handler.py
```diff
@@ -80,71 +57,47 @@ def change_loom_mode(
...
-                logger.debug(
-                    f"Current mode: {self.active_looms[abs_loom_file_path].get_connection().mode}, wanted mode {mode}"
-                )
-
-                if self.active_looms[abs_loom_file_path].get_connection().mode == mode:
-                    loom = self.active_looms[abs_loom_file_path]
+        with self.file_locks[abs_loom_file_path]:
+            if not os.path.exists(abs_loom_file_path):
+                logger.error(f"The file {loom_file_path} does not exists.")
+                raise ValueError(f"The file located at {abs_loom_file_path} does not exist.")
```
dweemx, line 83: Shouldn't this check be outside of the context manager ? Doesn't make sense to make a lock on a file that doesn't exist

opt/scopeserver/dataserver/utils/loom_file_handler.py
```diff
@@ -80,71 +57,47 @@ def change_loom_mode(
...
-        if abs_loom_file_path in self.active_looms:
-            logger.debug("Should be preloaded")
-            try:
-                logger.debug(
-                    f"Current mode: {self.active_looms[abs_loom_file_path].get_connection().mode}, wanted mode {mode}"
-                )
-
-                if self.active_looms[abs_loom_file_path].get_connection().mode == mode:
-                    loom = self.active_looms[abs_loom_file_path]
+        with self.file_locks[abs_loom_file_path]:
```
dweemx, line 80: Nice!

opt/scopeserver/dataserver/utils/loom.py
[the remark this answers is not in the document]
MaybeJustJames: You'll have to excuse my ignorance but I'm not seeing a sanity check here. I just see a NO-OP: get a clustering but don't use it.

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -31,9 +31,8 @@
 
 
 class Loom:
-    def __init__(self, partial_md5_hash, file_path, abs_file_path, loom_connection, loom_file_handler):
+    def __init__(self, file_path, abs_file_path, loom_connection, loom_file_handler):
```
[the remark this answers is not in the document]
MaybeJustJames: Why was this resolved? Can you add typings please?

opt/scopeserver/dataserver/utils/loom.py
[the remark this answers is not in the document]
MaybeJustJames: Sorry, I meant the exception

opt/scopeserver/dataserver/utils/loom_file_handler.py
```diff
@@ -80,71 +57,47 @@ def change_loom_mode(
...
-            except Exception as e:
-                logger.debug("Couldn't remove pickle SS")
-                logger.debug(e)
-                pass
         return self.active_looms[abs_file_path].get_connection()
 
-    def get_loom_absolute_file_path(self, loom_file_path: str) -> str:
-        return os.path.join(self.loom_dir, loom_file_path)
+    def get_loom_absolute_file_path(self, loom_file_path: Path) -> Path:
+        return self.loom_dir / loom_file_path
```
MaybeJustJames, line 66: This function [isn't really necessary](https://docs.python.org/3/library/pathlib.html#pathlib.Path.resolve)

opt/scopeserver/dataserver/utils/loom_file_handler.py
```diff
@@ -80,71 +57,47 @@ def change_loom_mode(
...
         self.global_looms = [x for x in os.listdir(self.loom_dir) if not os.path.isdir(os.path.join(self.loom_dir, x))]
 
-    def get_loom_connection(self, loom_file_path: str, mode: str = "r"):
+    def get_loom_connection(self, loom_file_path: Path, mode: str = "r") -> LoomConnection:
         logger.debug(f"Getting connection for {loom_file_path} in mode {mode}")
         return self.get_loom(loom_file_path=loom_file_path, mode=mode).get_connection()
 
-    def get_loom(self, loom_file_path: str, mode: str = "r"):
+    def get_loom(self, loom_file_path: Path, mode: str = "r") -> Loom:
         abs_loom_file_path = self.get_loom_absolute_file_path(loom_file_path)
```
MaybeJustJames, line 79:
```suggestion
        loom_file_path.resolve(strict=True)
```

opt/scopeserver/dataserver/utils/loom_file_handler.py
```diff
@@ -12,65 +16,38 @@
...
-    def load_loom_file(self, partial_md5_hash: str, file_path: str, abs_file_path: str, mode: str = "r"):
+    def load_loom_file(self, file_path: Path, abs_file_path: Path, mode: str = "r") -> Optional[Loom]:
         try:
             loom_connection = lp.connect(abs_file_path, mode=mode, validate=False)
+            return self.add_loom(file_path=file_path, abs_file_path=abs_file_path, loom_connection=loom_connection,)
         except KeyError as e:
             logger.error(e)
             os.remove(file_path)
             logger.warning(f"Deleting malformed loom {file_path}")
+            self.file_locks[abs_file_path].release()
```
MaybeJustJames, line 37: Why is this here? Seems like it's going to trigger a [RuntimeError](https://docs.python.org/3.8/library/threading.html#threading.Lock.release) when the context manager tries to release it.

opt/scopeserver/dataserver/utils/loom_file_handler.py
MaybeJustJames: I think the check should be after the file is locked. That way you make the operations _atomic_ and you avoid a potential data race:
1. Thread A checks `loom0` exists
1. Thread A locks `loom0 mutex`
1. Thread B checks `loom0` exists
1. Thread A deletes `loom0`
1. Thread A yields
1. Thread B locks `loom0 mutex`
1. OOPS! Thread B thinks `loom0` exists, but Thread A already deleted it

opt/scopeserver/dataserver/utils/loom_file_handler.py
```diff
@@ -80,71 +57,47 @@ def change_loom_mode(
...
-            logger.debug("Should be preloaded")
-            try:
-                logger.debug(
-                    f"Current mode: {self.active_looms[abs_loom_file_path].get_connection().mode}, wanted mode {mode}"
-                )
-
-                if self.active_looms[abs_loom_file_path].get_connection().mode == mode:
-                    loom = self.active_looms[abs_loom_file_path]
+        with self.file_locks[abs_loom_file_path]:
+            if not os.path.exists(abs_loom_file_path):
```
MaybeJustJames, line 81:
```suggestion
            if not abs_loom_file_path.exists():
```

opt/scopeserver/dataserver/utils/loom_file_handler.py
```diff
@@ -80,71 +57,47 @@ def change_loom_mode(
...
+                loom = self.active_looms[abs_loom_file_path]
+                try:
                     logger.debug(
-                        f"Returning pre-loaded loom file {loom_file_path}. Hash {partial_md5_hash}, object {id(loom)}"
+                        f"Current mode: {self.active_looms[abs_loom_file_path].get_connection().mode}, wanted mode {mode}"
                     )
-                    return loom
-            except AttributeError:
-                logger.error("Loom was previously closed")
+                    if self.active_looms[abs_loom_file_path].get_connection().mode == mode:
```
MaybeJustJames, line 91: What happens if the `mode` does not match? The `try`/`catch` falls through and you return the loom at the end of the function

opt/scopeserver/dataserver/utils/loom.py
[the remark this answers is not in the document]
KrisDavie: I see what you're saying. The sanity check was relying on the fact that the getter will fail if the ID doesn't exist in the file. However, we can make this a more explicit check rather than relying on the failure.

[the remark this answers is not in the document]
MaybeJustJames: Yes please, especially since there's nothing in the definition of `get_clustering_by_id` to indicate that it might fail or how it might fail. I presume now that it's with an `IndexError`?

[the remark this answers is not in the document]
KrisDavie: Indeed, I will rewrite it to check the presence of the index specifically 👍

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -43,37 +43,56 @@ def __init__(self, partial_md5_hash, file_path, abs_file_path, loom_connection,
...
+            logger.debug(f"Building mmus Search Spaces for {self.file_path}")
+            self.mmus_ss = ss.SearchSpace(loom=self, cross_species="mmus").build()
+        logger.debug(f"Building self Search Spaces for {self.file_path}")
+
+        self.ss = ss.SearchSpace(loom=self).build()
+        self.write_ss()
+
+    def write_ss(self) -> None:
+        self.ss.loom = None  # Remove loom connection to enable pickling
+        logger.debug(f"Built all Search Spaces for {self.file_path}")
```
KrisDavie, line 75: A copy of the ss should be made here, to prevent an issue with the dictionary being changed during writing. Importantly, the dict.copy() function should NOT be used, as it does not copy the extra attributes we have on the SearchSpace class. The copy module should be used to copy the whole object.

opt/scopeserver/dataserver/utils/loom_file_handler.py
KrisDavie: This function takes in loom file names which are not a full path object, which means we need to build the path still. Maybe `get_loom_absolute_file_path` isn't the best name

opt/scopeserver/dataserver/utils/loom_file_handler.py
KrisDavie: Yeah. I;'m not sure if theres something better here, we should at least log the error. If we don't return the loom, the frontend could end up hanging somewhere, but we don't have feedback to the user that something failed.

opt/scopeserver/dataserver/utils/loom_file_handler.py
KrisDavie: Yeah, you're right. I forgot that the below return would still allow the contect manager to rexit gracefully, releasing the lock itself.

opt/scopeserver/dataserver/utils/loom_file_handler.py
KrisDavie: To be clearer, the frontend often send just the name of a loom file, which we need to build the path from, based on the knowledge of where the backend is storing these files

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -31,9 +34,10 @@
 
 
 class Loom:
-    def __init__(self, partial_md5_hash, file_path, abs_file_path, loom_connection, loom_file_handler):
+    def __init__(
+        self, file_path: Path, abs_file_path: Path, loom_connection: LoomConnection, loom_file_handler,
```
KrisDavie, line 38: I'm not sure how to resolved the typing for loom_file_handler here. To use the `LoomFileHandler` class, we need to import the `loom_file_handler` module, which imports the `Loom` class from the `loom` module. This ends up being circular and everything breaks.

opt/scopeserver/dataserver/utils/loom_file_handler.py
KrisDavie: See above comment

opt/scopeserver/dataserver/utils/loom_file_handler.py
MaybeJustJames: OK I see

opt/scopeserver/dataserver/utils/loom.py
dweemx: They recommend refactoring the code to avoid these circular dependencies (probably the best). Otherwise you could use this trick: https://www.stefaanlippens.net/circular-imports-type-hints-python.html

KrisDavie: As discussed on Skype, this refactoring will come _after_ we decide on the new format for the loom files, as this is going to require a lot of work here also.

opt/scopeserver/dataserver/utils/loom_file_handler.py
KrisDavie: As discussed on Slack, this will be done in a separate PR for full integration of the Result class.

MaybeJustJames:
The state of this branch is wildly confusing. Could you please rebase on master?

opt/scopeserver/dataserver/utils/loom_file_handler.py
```diff
@@ -80,71 +56,51 @@ def change_loom_mode(
...
+                        logger.debug(f"Returning pre-loaded loom file {loom_file_path}. Object {id(loom)}")
+                        return loom
+                    else:
+                        logger.error(
+                            f"Mode {mode} was requested for {loom_file_path}, but mode is currently {self.active_looms[abs_loom_file_path].get_connection().mode}"
+                        )
```
MaybeJustJames, line 96: Maybe add a `loom = None` here?

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -43,36 +47,60 @@ def __init__(self, partial_md5_hash, file_path, abs_file_path, loom_connection,
...
+        self.load_ss(self.ss_pickle_name)
 
+    def load_ss(self, ss_pickle_name: Path) -> None:
         try:
-            with open(self.ss_pickle_name, "rb") as fh:
-                logger.debug(f"Loading prebuilt SS for {file_path} from {self.ss_pickle_name}")
+            with open(ss_pickle_name, "rb") as fh:
+                logger.debug(f"Loading prebuilt SS for {self.file_path} from {ss_pickle_name}")
                 self.ss = pickle.load(fh)
-                if self.ss.search_space_version < ss.CURRENT_SS_VERISON:
```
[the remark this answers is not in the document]
KrisDavie: The code just below this takes this place. This was the little bit that I mentioned somehow got in during the rebase fix.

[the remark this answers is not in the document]
MaybeJustJames: Ah ok cool.

opt/scopeserver/dataserver/utils/loom_file_handler.py
KrisDavie: This is something to decide on, _if_ `loom = None` works, it will result in the loom disappearing for the frontend user. Ideally, we actually want to respond with an error to let them know that it failed, but this is the bit w discussed about putting it in a different PR with the result class.

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -43,36 +47,60 @@ def __init__(self, partial_md5_hash, file_path, abs_file_path, loom_connection,
...
+                try:
+                    assert self.ss.search_space_version == ss.CURRENT_SS_VERISON
+                except AssertionError:
+                    logger.error(
+                        f"Cached search space version {self.ss.search_space_version} is not {ss.CURRENT_SS_VERISON}. Rebuilding search space..."
+                    )
+                    self.build_ss()
+                except AttributeError:
+                    logger.error(f"Search space has no version key and is likely legacy. Rebuilding search space...")
+                    self.build_ss()
```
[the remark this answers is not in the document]
KrisDavie: I'll look into moving them over.

I did debate having this error as a single one, but wanted to be able to report the old search space version, which I can't do in one error without 2 if's (on legacy search spaces (no version) your code will fail on the logger because `self.ss.search_space_version` doesn't exit).

Are you insistent on this being collapsed into 1 error?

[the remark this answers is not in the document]
MaybeJustJames: No not fixed on that at all. I would prefer you didn't use exceptions for flow control though.

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -431,7 +384,7 @@ def add_user_clustering(self, request) -> Tuple[bool, str]:
         loom.attrs["MetaData"] = json.dumps(metaJson)
         self.loom_connection = self.lfh.change_loom_mode(self.file_path, mode="r")
 
-        self.update_ss(update="clusterings")
+        ss.update_ss(self, update="clusterings")
```
MaybeJustJames, line 387: Should this not be `self.ss.update(`?

opt/scopeserver/dataserver/utils/search_space.py
```diff
@@ -182,3 +184,53 @@ def add_metrics(self) -> None:
         for metric in self.meta_data["metrics"]:
             metrics.append(metric["name"])
         self.add_elements(elements=metrics, element_type="metric")
+
+
+def load_ss(loom) -> SearchSpace:
```
MaybeJustJames, line 189: Maybe just `load` now?

opt/scopeserver/dataserver/utils/search_space.py
```diff
@@ -182,3 +184,53 @@ def add_metrics(self) -> None:
...
+    ss.loom = None  # Remove loom connection to enable pickling
+    ss.dfh = None
+    cur_ss = deepcopy(ss)
+    with open(loom.ss_pickle_name, "wb") as fh:
+        logger.debug(f"Writing SS for {loom.file_path} to {loom.ss_pickle_name}")
+        pickle.dump(cur_ss, fh)
+
+
+def update_ss(loom, update: str) -> SearchSpace:
+    ss = loom.ss
```
MaybeJustJames, line 228: Aaaahhh!

KrisDavie:
Yes, was still working on it. The only open question is the one above about returning None rather than Loom. I'm inclined to leave it as is for now until we implement the whole result bit.

opt/scopeserver/dataserver/utils/search_space.py
MaybeJustJames: Does the `SearchSpace` _really_ need to know about the loom? The loom should know about the search space but not the other way around?

PR #361, opened 2020-06-22
Feature/356 link out to gprofiler

dweemx:


opt/scopeserver/dataserver/modules/gserver/GServer.py
```diff
@@ -987,6 +988,65 @@ def loomUploaded(self, request, content):
...
+            lambda x: functools.reduce(np.logical_and, ~np.isnan(x)), axis=1
+        )
+        return cluster_marker_metrics[nonan_marker_mask]
+
+    def getGProfilerLink(self, request, context):
+        cluster_marker_table = self.get_cluster_marker_table(
+            loomFilePath=request.loomFilePath, clusteringID=request.clusteringID, clusterID=request.clusterID
+        )
+        if "avg_logFC" not in cluster_marker_table.columns:
+            return s_pb2.GProfilerLinkOutReply(url="Error")
```
KrisDavie, line 1018: In some of the later functions, I've been using a (Response: bool, Message: str) format to return any errors, and whether the function worked. It might be nice to do the same here, then you can return a meaningful error message to the user. 

Either way it would be nice to log an error on the backend here.

src/components/common/API.jsx
```diff
@@ -1146,6 +1146,41 @@ class API {
...
+        organism,
+        token
+    ) {
+        let query = {
+            loomFilePath: this.getActiveLoom(),
+            clusteringID: clusteringID,
+            clusterID: clusterID,
+            topNumFeatures: topNumFeatures,
+            organism: token !== null ? '' : organism,
+            token: token === null ? '' : token,
```
KrisDavie, line 1167: This could just be 
```suggestion
            organism: token !== null ? token : organism,
```
Then you wouldn't need the extra option in the message, and you wouldn't need to do the check on the backend, the `token` gets passed via `organism` for the request anyway

opt/scopeserver/dataserver/modules/gserver/GServer.py
```diff
@@ -987,6 +988,65 @@ def loomUploaded(self, request, content):
...
+            )
+
+        md_clustering = loom.get_meta_data_clustering_by_id(id=clusteringID, secret=self.config["dataHashSecret"])
+        md_cmm = md_clustering["clusterMarkerMetrics"]
+        cluster_marker_metrics = merge_cluster_marker_metrics(metrics=[create_cluster_marker_metric(x) for x in md_cmm])
+        # Keep only non-zeros elements
+        nonan_marker_mask = cluster_marker_metrics.apply(
+            lambda x: functools.reduce(np.logical_and, ~np.isnan(x)), axis=1
+        )
+        return cluster_marker_metrics[nonan_marker_mask]
```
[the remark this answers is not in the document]
KrisDavie: Is there not a lot of overlap/duplicated code here with `getMarkerGenes`? 

https://github.com/aertslab/SCope/blob/041bf6b56ed3e67f071b86e745617711e2c0b2f9/opt/scopeserver/dataserver/modules/gserver/GServer.py#L595-L639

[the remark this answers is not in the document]
dweemx: Yeah true, I'll abstract this into the `loom.py` file

opt/scopeserver/dataserver/modules/gserver/GServer.py
dweemx: I was lazy here :p but indeed better to have some structure for returning an error

KrisDavie:
Now I'm actually playing with it. I realised that all of the information you're getting on the backend has already been sent to the frontend. Is it not possible to do this link construction on the frontend only? This might then be able to integrate with a filtering of the marker gene table.

KrisDavie:
Also, selecting all of the options (Top100-Top500) fails, removing top500 gives a "URI is too long error"

dweemx:
> Also, selecting all of the options (Top100-Top500) fails, removing top500 gives a "URI is too long error"
> 
> Selecting 400 + 500 only still gives URI error, one or the other alone works

Yeah also noticed this... I'll have to constrain to 100 + 200 + 300 genes

dweemx:
> Now I'm actually playing with it. I realised that all of the information you're getting on the backend has already been sent to the frontend. Is it not possible to do this link construction on the frontend only? This might then be able to integrate with a filtering of the marker gene table.

It's just much more convenient to do the processing in Python I find

MaybeJustJames:
What is gProfiler?

dweemx:
> What is gProfiler?

A web-server to perform functional enrichment analysis of a list of genes

package.json
```diff
@@ -97,9 +97,11 @@
         "@babel/core": "^7.9.6",
         "@babel/plugin-proposal-class-properties": "^7.8.3",
         "@babel/plugin-proposal-object-rest-spread": "^7.9.6",
+        "@babel/plugin-transform-runtime": "^7.10.3",
```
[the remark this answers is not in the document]
dweemx: Need this otherwise `async` cannot be interpreted

opt/scopeserver/dataserver/modules/gserver/GServer.py
```diff
@@ -987,6 +973,46 @@ def loomUploaded(self, request, content):
         uploadedLooms[request.UUID].add(request.filename)
         return s_pb2.LoomUploadedReply()
 
+    def getGProfilerLink(self, request, context):
```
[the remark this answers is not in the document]
dweemx: Yes this the RPC class method that the client calls to get the GProfiler link. Where would you put it ?

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,335 @@
+import _ from 'lodash';
```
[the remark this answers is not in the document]
dweemx: Indeed this should be removed

src/components/common/API.jsx
```diff
@@ -1146,6 +1146,40 @@ class API {
...
+            topNumFeatures: topNumFeatures,
+            organism: token !== null ? token : organism,
+        };
+        console.log('Run getGProfilerLink');
+        console.log(query);
+
+        return new Promise((resolve, reject) => {
+            this.getConnection().then((gbc) => {
+                if (DEBUG) console.log('getGProfilerLink', query);
+                gbc.services.scope.Main.getGProfilerLink(
```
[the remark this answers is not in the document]
dweemx: > Ah, here's where the link is created on the server? Why are you using a raw Promise in an `async` function?

I understand but then how would the Promise resolve since the result is nested ?

> Why not create the link here? Do you not have the information you need?

Data is there but processing the data is much easier too me on the back-end in Python (could be biased)

opt/scopeserver/dataserver/modules/gserver/GServer.py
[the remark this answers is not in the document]
KrisDavie: I think James means in the backend rather than just building it on the frontend

opt/scopeserver/dataserver/modules/gserver/GServer.py
```diff
@@ -607,35 +608,20 @@ def getMarkerGenes(self, request, context):
...
-            def merge_cluster_marker_metrics(metrics):
-                return functools.reduce(
-                    lambda left, right: pd.merge(left, right, left_index=True, right_index=True, how="outer"), metrics,
+                    values=cluster_marker_metrics[metric["accessor"]],
                 )
 
-            cluster_marker_metrics = merge_cluster_marker_metrics(
-                metrics=[create_cluster_marker_metric(x) for x in md_cmm]
+            cluster_marker_metrics = loom.get_cluster_marker_table(
+                clusteringID=request.clusteringID, clusterID=request.clusterID
```
KrisDavie, line 620: This call eventually leads to calling `get_meta_data_clustering_by_id` which requires the config's dataHashSecret to validate Annotations. Not passing it down the chain means that the backend bails out when getting the marker table.
```suggestion
            cluster_marker_metrics = loom.get_cluster_marker_table(
                clusteringID=request.clusteringID, clusterID=request.clusterID, secret=self.config['dataHashSecret']
```

opt/scopeserver/dataserver/modules/gserver/GServer.py
```diff
@@ -987,6 +973,46 @@ def loomUploaded(self, request, content):
...
+            logger.info(error_message)
+            return s_pb2.GProfilerLinkOutReply(error=s_pb2.ErrorReply(type="Value Error", message=error_message,))
+
+        def get_top_marker_genes_by(n, by="avg_logFC"):
+            top_genes_ranked = cluster_marker_table.sort_values(by=by, ascending=False).head(n=n).index.values
+            return np.insert(top_genes_ranked, 0, f">Top_{n}")
+
+        top_ranked_gene_lists = [
+            get_top_marker_genes_by(n=topNumFeaturesValue) for topNumFeaturesValue in sorted(request.topNumFeatures)
+        ]
```
KrisDavie, line 992: There could be a check here to confirm that all of the selected markers are in the same direction. 

It would actually be nice to add a separation between up and down regulated genes, much like you do for top100, top200 etc.

Just selecting the "top" X genes, could actually lead to a selection of genes with both positive and negative logFC values. From a biological view this could dilute the results as the upregulated and downregulated genes may have different functions.

src/components/common/API.jsx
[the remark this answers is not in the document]
KrisDavie: If we did decide to do this fully on the frontend, the ReactTable module could be used to filter the data, and then this data could be extracted (https://spectrum.chat/react-table/general/best-way-to-get-currently-visible-data~3bc65318-678d-4f7e-88f5-7c5050668287). In this case, we don't need to do anything to deal with the data as we just get the gene list directly, people would also have a better refinement on which genes they are submitting.

Although this would mean submitting multiple sets (as is done now) would be difficult.

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -976,3 +977,23 @@ def get_cluster_marker_metrics(self, clustering_id: int, cluster_id: int, metric
             cluster_marker_metric, index=self.get_genes()[marker_mask], columns=[metric_accessor]
         )
         return cluster_marker_metric_df
+
+    def get_cluster_marker_table(self, clusteringID, clusterID) -> pd.DataFrame:
```
KrisDavie, line 981:
```suggestion
    def get_cluster_marker_table(self, clusteringID, clusterID, secret) -> pd.DataFrame:
```

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -976,3 +977,23 @@ def get_cluster_marker_metrics(self, clustering_id: int, cluster_id: int, metric
...
+            return self.get_cluster_marker_metrics(
+                clustering_id=clusteringID, cluster_id=clusterID, metric_accessor=metric["accessor"]
+            )
+
+        def merge_cluster_marker_metrics(metrics):
+            return functools.reduce(
+                lambda left, right: pd.merge(left, right, left_index=True, right_index=True, how="outer"), metrics,
+            )
+
+        md_clustering = self.get_meta_data_clustering_by_id(id=clusteringID)
```
KrisDavie, line 992:
```suggestion
        md_clustering = self.get_meta_data_clustering_by_id(id=clusteringID, secret=secret)
```

opt/scopeserver/dataserver/modules/gserver/GServer.py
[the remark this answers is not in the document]
MaybeJustJames: Well, as you say below, you have all the data in the client to generate the link _in the client_. An API call is the worst case scenario see [falacies of distributed computing](https://en.wikipedia.org/wiki/Fallacies_of_distributed_computing).
I request that you _try_ to generate the link on the client. If it's too much we can pair and try to solve it. If, after that it's still to much then we can generate it on the server.

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,335 @@
...
+                        }
+                        {this.state.gProfilerURL !== null && (
+                            <Button
+                                type='button'
+                                value='goto-gprofiler'
+                                onClick={handleClickGotoGProfilerURL}
+                                primary>
+                                {'Go to g:Profiler'}
+                            </Button>
+                        )}
```
KrisDavie, line 327: This might be nicer if the button changes to the the one for the link, rather than a second button appearing, a little more intuitive maybe.

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,335 @@
...
+    Modal,
+    Form,
+    Checkbox,
+    Table,
+    Input,
+    Select,
+    Label,
+} from 'semantic-ui-react';
+import { instanceOf } from 'prop-types';
+import { withCookies, Cookies } from 'react-cookie';
```
KrisDavie, line 17: You're not actually using cookies here are you? All the cookie code could be removed if not.

package.json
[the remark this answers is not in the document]
MaybeJustJames: `async` has been supported in browsers [for ages](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function#Browser_compatibility).
Chrome: 2016
Firefox: 2017
Opera (based on chromium): 2016
Edge: 2017 (now also based on chromium)
Safari: 2016

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,335 @@
...
+            this.setState({
+                gProfilerURL: gProfilerLinkResponse.url,
+            });
+        };
+
+        const organisms = [
+            { key: 1, text: 'Human', value: 'hsapiens' },
+            { key: 2, text: 'Mouse', value: 'mmusculus' },
+            { key: 3, text: 'Drosophila melanogaster', value: 'dmelanogaster' },
+        ];
```
KrisDavie, line 150: gProfiler supports more organisms than this and they have an API for getting that list from them (See: https://biit.cs.ut.ee/gprofiler/page/apis)

It would be nice if we could allow all of these organisms to work.

Regardless, we should use either full species names (Homo Sapeins, Mus Musculus, Drosophila Melanogaster) OR common names (Human, Mouse, Fly) rather than a combination (I prefer Species names).

src/components/common/API.jsx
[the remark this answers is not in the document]
MaybeJustJames: See comment above

src/components/common/ViewerSidebar.jsx
```diff
@@ -1225,6 +1226,13 @@ class ViewerSidebar extends Component {
                             {markerTable}
                             {legendTable}
                             {downloadSubLoomButton()}
+                            <GProfilerModal
+                                numFeatures={md.genes.length}
+                                clusteringID={
+                                    activeFeatures[i].metadata.clusteringID
+                                }
+                                clusterID={activeFeatures[i].metadata.clusterID}
+                            />
```
KrisDavie, line 1235: There is no check to ensure that we're looking at a clustering, or if there are any marker genes. This currently causes the button to appear on all searches and when there are no marker genes, can crash the frontend.

`md.genes.length` shouldn't be called without confirming that `md.genes` isn't null.

opt/scopeserver/dataserver/modules/gserver/GServer.py
KrisDavie: Also, we use `avg_logFC`, but not everyone might, using the frontend data and allowing sorting/filtering would prevent us limiting the looms that this can be used for.

KrisDavie: Further to this actually, this could also be useful for analysing target genes from regulons, but these don't have a log_fc

package.json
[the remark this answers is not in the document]
dweemx: @MaybeJustJames Here is the detailed reason why I added: https://andrew-flower.com/blog/Async-Await-with-React

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -976,3 +977,23 @@ def get_cluster_marker_metrics(self, clustering_id: int, cluster_id: int, metric
...
+
+    def get_cluster_marker_table(self, clusteringID, clusterID) -> pd.DataFrame:
+        def create_cluster_marker_metric(metric):
+            return self.get_cluster_marker_metrics(
+                clustering_id=clusteringID, cluster_id=clusterID, metric_accessor=metric["accessor"]
+            )
+
+        def merge_cluster_marker_metrics(metrics):
+            return functools.reduce(
+                lambda left, right: pd.merge(left, right, left_index=True, right_index=True, how="outer"), metrics,
```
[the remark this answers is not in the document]
dweemx: > This `lambda` is what should be in the function, not the `reduce`.

I don't get your comment; you need the reduce function in order to merge the different `pd.DataFrame` together right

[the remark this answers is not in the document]
dweemx: Oh I see you put the reduce function at higher level. But then I don't see the difference between the 2 approaches... It should give the same results

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -976,3 +977,23 @@ def get_cluster_marker_metrics(self, clustering_id: int, cluster_id: int, metric
...
+            )
+
+        md_clustering = self.get_meta_data_clustering_by_id(id=clusteringID)
+        md_cmm = md_clustering["clusterMarkerMetrics"]
+        cluster_marker_metrics = merge_cluster_marker_metrics(metrics=[create_cluster_marker_metric(x) for x in md_cmm])
+        # Keep only non-zeros elements
+        nonan_marker_mask = cluster_marker_metrics.apply(
+            lambda x: functools.reduce(np.logical_and, ~np.isnan(x)), axis=1
+        )
+        return cluster_marker_metrics[nonan_marker_mask]
```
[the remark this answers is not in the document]
dweemx: Basically I want to get a mask based on the non NaN values in each of the columns of the `pd.DataFrame`. So my `nonan_marker_mask` should be a 1d array.
Could be a shortcut what you're proposing but didn't test it

src/components/common/API.jsx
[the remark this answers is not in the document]
dweemx: I have a feeling that it's going to be harder in JS but I can try

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,335 @@
...
+            ...INITIAL_STATE,
+            showModal: true,
+        });
+    };
+
+    closeModal = () => {
+        this.setState({ showModal: false });
+    };
+
+    render() {
```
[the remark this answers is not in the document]
dweemx: Yeah I'll simplify this

src/components/common/GProfilerModal.tsx
dweemx: Yeah I know but I think for now we constrain it to these main species no ?

src/components/common/ViewerSidebar.jsx
dweemx: Indeed, good catch

src/components/common/GProfilerModal.tsx
dweemx: Yeah that's right, will remove

src/components/common/GProfilerModal.tsx
dweemx: Not bad idea, I'll change it to your suggestion

src/components/common/GProfilerModal.tsx
MaybeJustJames: Because there's no rush. I'd prefer this be done "properly" the first time.

package.json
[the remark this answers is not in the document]
MaybeJustJames: This is surprising. Do you observe a `ReferenceError: regeneratorRuntime is not defined` when you run `async`/`await` code without this plugin? If so, what browser and browser version are you running?

opt/scopeserver/dataserver/utils/loom.py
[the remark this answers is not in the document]
MaybeJustJames: Yes it will give the same result, but we're writing for people not the Python interpreter.

opt/scopeserver/dataserver/utils/loom.py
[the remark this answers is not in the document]
MaybeJustJames: Could you check it please. Best to keep numpy logic as simple and as "inside" numpy as possible.

src/components/common/API.jsx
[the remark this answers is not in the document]
MaybeJustJames: Sure it might be harder. Let me know if you want to pair.

package.json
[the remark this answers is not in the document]
dweemx: Nope no errors; the plugin does solve this issue
Version: `Mozilla Firefox 78.0.1`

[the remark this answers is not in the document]
MaybeJustJames: Sorry Max I don't understand.
> no errors

Then what is the issue the plugin solves?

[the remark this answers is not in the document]
dweemx: Sorry I misread your comment.
So without adding this plugin I indeed observe the error you're mentioning `ReferenceError: regeneratorRuntime is not defined`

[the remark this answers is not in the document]
MaybeJustJames: Ah ok. Better keep it then. Babel is a complicated beast. I'm very curious what is going on though.

opt/scopeserver/dataserver/modules/gserver/GServer.py
```diff
@@ -18,6 +18,7 @@
 import requests
 import uuid
 import functools
+from urllib.parse import urlencode
```
KrisDavie, line 21: This isn't used now

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,465 @@
...
+                                    <Form.Field
+                                        control={Select}
+                                        label='Sort Features By'
+                                        options={this.availableSortBy}
+                                        placeholder='Sort By'
+                                        onChange={this.handleSelectSortBy}
+                                        value={selectedSortBy}
+                                    />
+                                    <Form.Field
+                                        control={Select}
```
KrisDavie, line 409: This would be very nice as an `Dropdown` instead of a `Select`, this would let people type to filter the list.

opt/scopeserver/dataserver/modules/gserver/GServer.py
```diff
@@ -18,6 +18,7 @@
 import requests
 import uuid
 import functools
```
KrisDavie, line 20: This isn't used in here now

opt/scopeserver/dataserver/utils/loom.py
```diff
@@ -976,3 +977,19 @@ def get_cluster_marker_metrics(self, clustering_id: int, cluster_id: int, metric
             cluster_marker_metric, index=self.get_genes()[marker_mask], columns=[metric_accessor]
         )
         return cluster_marker_metric_df
+
+    def get_cluster_marker_table(self, clustering_id: int, cluster_id: int, secret: str) -> pd.DataFrame:
```
KrisDavie, line 981: Nice!

src/components/common/API.jsx
```diff
@@ -1146,6 +1146,10 @@ class API {
     getColors() {
         return this.colors;
     }
+
```
KrisDavie, line 1149: This comment isn't needed?

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,465 @@
...
+            measure_underrepresentation: 'false',
+            domain_scope: 'annotated',
+            significance_threshold_method: 'g_SCS',
+            user_threshold: '0.05',
+            numeric_namespace: 'ENTREZGENE_ACC',
+            sources: 'GO:MF,GO:CC,GO:BP,KEGG,TF,REAC,MIRNA,HPA,CORUM,HP,WP',
+            background: '',
+        };
+
+        var gProfilerQueryString = Object.keys(gProfilerQueryData)
```
KrisDavie, line 189: Should be const

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,465 @@
...
+                    acc +
+                    `>Top_${el.numFeatures}` +
+                    '\n' +
+                    el.topSortedFeatures +
+                    '\n'
+                );
+            }, '');
+    };
+
+    buildGProfilerLink = (organism: string, query: string) => {
```
KrisDavie, line 172: Rather than have two buttons, wouldn't it make sense for there to be one button, which first builds the link and then opens the URL? I can't think of a reason when a user would want to make the URL without visiting.

All of the logic would remain the same, the same checks etc can be performed, but for the user it is nicer to only press one button.

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,465 @@
...
+            this.setState({
+                error: 'Please select an organism',
+                gProfilerURL: null,
+            });
+            return;
+        }
+        if (topNumFeatures.length == 0) {
+            this.setState({
+                error:
+                    'No gene list selected. One gene list is at least required. ',
```
KrisDavie, line 222:
```suggestion
                    'No gene list selected. At least one gene list is required. ',
```

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,465 @@
...
+                gProfilerURL: null,
+            });
+            return;
+        }
+        if (topNumFeatures.length == 0) {
+            this.setState({
+                error:
+                    'No gene list selected. One gene list is at least required. ',
```
KrisDavie, line 230:
```suggestion
                    'No gene list selected. At least one gene list is required. ',
```

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,465 @@
...
+
+    render() {
+        const {
+            showModal,
+            selectedOrganism,
+            selectedSortBy,
+            gProfilerToken,
+        } = this.state;
+
+        console.log(this.state);
```
KrisDavie, line 295: Remove logging

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,465 @@
...
+            showModal: true,
+        });
+    };
+
+    closeModal = () => {
+        this.setState({ showModal: false });
+    };
+
+    handleOpenModal = () => {
+        console.log(this.state);
```
KrisDavie, line 85: Remove logging

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,465 @@
...
+                    key: idx + 1,
+                    text: availableOrganism['display_name'],
+                    value: availableOrganism['id'],
+                };
+            })
+            .sort((a, b) => (a['text'] < b['text'] ? -1 : 1));
+        this.setState({ availableOrganisms: _availableOrganisms });
+    }
+
+    async fetchAvailableOrganisms() {
```
KrisDavie, line 275: If this fails for some reason, is there a good way to present a message to the user? Also if this fails, gProfiler is likely down, so it might be a good idea to disable the button, and show a message that is it down.

src/components/common/API.jsx
dweemx: No indeed

src/components/common/GProfilerModal.tsx
dweemx: Yeah makes sense, I'll make it one

opt/scopeserver/dataserver/utils/loom.py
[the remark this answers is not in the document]
MaybeJustJames: If it is I'd still like to try to rework things such that the logic is all happening "inside" numpy/pandas. (As much as I appreciate you reaching for the `reduce` function ;))

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,462 @@
...
+    gProfilerToken: null,
+};
+
+class GProfilerPopup extends Component<
+    IGProfilerPopupProps & RouteComponentProps,
+    IGProfilerPopupState
+> {
+    featureMetadata: any;
+    featureMetricTable: any;
+    availableSortBy: object[];
```
MaybeJustJames, line 47: Can these be made `private readonly` and the types be made more specific please? They may not even be necessary, e.g. `availableSortBy` could be derived in a reducer.

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,462 @@
...
+            (metric, idx) => {
+                return {
+                    key: idx,
+                    text: metric.name,
+                    value: metric.accessor,
+                };
+            }
+        );
+        this.featureMetricTable = this.buildMetricTable();
+        this.handleOpenModal.bind(this);
```
MaybeJustJames, line 63: Are all of these binds really necessary? What is the `this` context?

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,462 @@
...
+        this.handleCloseModal.bind(this);
+        this.handleSelectOrganism.bind(this);
+        this.handleSelectSortBy.bind(this);
+        this.handleChangeToken.bind(this);
+        this.handleClickGotoGProfilerURL.bind(this);
+    }
+
+    openModal = () => {
+        const { availableOrganisms, error, ...rest } = INITIAL_STATE;
```
MaybeJustJames, line 72: This line is a no-op.

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,462 @@
...
+        this.setState({
+            ...rest,
```
MaybeJustJames, line 74:
```suggestion
```

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,462 @@
...
+        } = this.state;
+
+        return (
+            <>
+                <Modal
+                    as={Form}
+                    trigger={
+                        <Button
+                            color='orange'
+                            onClick={this.handleOpenModal}
```
MaybeJustJames, line 311: Just `setState` directly here.

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,462 @@
...
+                            style={{
+                                marginTop: '10px',
+                                marginBottom: '10px',
+                                width: '100%',
+                            }}>
+                            Run g:Profiler Gene List Enrichment
+                        </Button>
+                    }
+                    onClose={this.handleCloseModal}
```
MaybeJustJames, line 320: Just `setState` directly here.

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,462 @@
...
+    };
+
+    handleClickCreateGProfilerLink = async () => {
+        const {
+            topNumFeatures,
+            gProfilerToken,
+            selectedOrganism,
+            selectedSortBy,
+        } = this.state;
+        if (selectedSortBy === null) {
```
MaybeJustJames, line 208: This should not be possible. Use typescript to ensure `selectedSortBy` is never null (a non-nullable string?) and use a default value in the UI control to ensure there's always a value selected.

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,462 @@
...
+            });
+            return null;
+        }
+        if (topNumFeatures.length == 0) {
+            this.setState({
+                error:
+                    'No gene list selected. At least one gene list is required.',
+            });
+            return null;
+        }
```
MaybeJustJames, line 237:
```suggestion
```

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,462 @@
...
+            topNumFeatures,
+            sortedFeatureMetricTable
+        );
+        const organism =
+            gProfilerToken !== null ? gProfilerToken : selectedOrganism;
+        const gProfilerLink = this.buildGProfilerLink(
+            organism,
+            topFeatureQuery
+        );
+        if (gProfilerLink.length > 8000) {
```
MaybeJustJames, line 252: Give the `8000` value a name: define it as a constant.

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,462 @@
...
+
+        const sortedFeatureMetricTable = this.featureMetricTable.sort(
+            (a, b) => b[selectedSortBy] - a[selectedSortBy]
+        );
+        const topFeatureQuery = this.buildFeatureQuery(
+            topNumFeatures,
+            sortedFeatureMetricTable
+        );
+        const organism =
+            gProfilerToken !== null ? gProfilerToken : selectedOrganism;
```
MaybeJustJames, line 247: Really?

src/components/common/ViewerSidebar.jsx
```diff
@@ -30,6 +30,7 @@ import { BackendAPI } from '../common/API';
 import Metadata from '../common/Metadata';
 import FileDownloader from '../../js/http';
 import CollaborativeAnnotation from './CollaborativeAnnotation';
+import GProfilerModal from './GProfilerModal.tsx';
```
MaybeJustJames, line 33:
```suggestion
import GProfilerModal from './GProfilerModal';
```
I would be happy to see this in `components/GProfiler/....`

src/components/common/ViewerSidebar.jsx
```diff
@@ -442,7 +443,8 @@ class ViewerSidebar extends Component {
                 let markerTable = '',
                     legendTable = '',
                     cellTypeAnnoTable = '',
-                    downloadSubLoomButton = () => '';
```
MaybeJustJames, line 445: This is not necessary

src/components/common/ViewerSidebar.jsx
```diff
@@ -1016,6 +1018,8 @@ class ViewerSidebar extends Component {
                             </Button>
                         </div>
                     );
+
+                    gProfilerModal = <GProfilerModal featureMetadata={md} />;
```
MaybeJustJames, line 1022: This is not necessary

src/components/common/ViewerSidebar.jsx
```diff
@@ -1225,6 +1229,7 @@ class ViewerSidebar extends Component {
                             {markerTable}
                             {legendTable}
                             {downloadSubLoomButton()}
+                            {gProfilerModal}
```
MaybeJustJames, line 1232:
```suggestion
                            <GProfilerModal featureMetadata={md} />
```

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,462 @@
...
+    Input,
+    Select,
+    Label,
+    Dropdown,
+} from 'semantic-ui-react';
+
+interface IGProfilerPopupState {
+    error: string;
+    showModal: boolean;
+    topNumFeatures: number[];
```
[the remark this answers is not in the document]
dweemx: `topNumFeatures` is array of number representing the top number of features to pick from the gene list in order to create `topNumFeautres.length` gene lists that will be send to g:Profiler.
You don't like the variable name, that's what you're saying ?

src/components/common/GProfilerModal.tsx
```diff
@@ -0,0 +1,462 @@
...
+    Form,
+    Checkbox,
+    Table,
+    Input,
+    Select,
+    Label,
+    Dropdown,
+} from 'semantic-ui-react';
+
+interface IGProfilerPopupState {
```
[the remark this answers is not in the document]
dweemx: I don't really agree that it should be in the Redux store: this state does not need to be shared to other component. Why you put something in the Redux store if other components of the App don't need to access this data ?

src/components/common/GProfilerModal.tsx
dweemx: This is a bad practice

src/components/common/GProfilerModal.tsx
dweemx: This is a bad practice

src/components/common/GProfilerModal.tsx
MaybeJustJames: "bad practice" is not useful reasoning. Why? What's bad about it? Under what circumstances?

src/components/common/GProfilerModal.tsx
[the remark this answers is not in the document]
MaybeJustJames: I'm not saying anything. I was just struggling to understand what it meant. I'd like to see what you wrote above in a doc comment please

src/components/common/GProfilerModal.tsx
MaybeJustJames: ~~reducer~~ selector.

src/components/common/GProfilerModal.tsx
dweemx: Using `() => setState({...})`, you will declare a new function on each component rendering

MaybeJustJames: OK so the implied concern is that we'll overload the GC? This model is performance sensitive? This isn't very convincing. I might be swayed by a "readability" argument

src/components/common/GProfilerModal.tsx
dweemx: Yes unfortunately, `this` is the component class context

src/components/common/ViewerSidebar.jsx
dweemx: I don't understand why this is not needed because the button is displayed by calling a function. Could you elaborate?

src/components/common/GProfilerModal.tsx
dweemx: You're suggesting no changes here...

src/components/common/ViewerSidebar.jsx
MaybeJustJames: I'm afraid I've lost the context for why I said this. Sorry. Ignore.

opt/scopeserver/dataserver/utils/loom.py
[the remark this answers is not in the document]
dweemx: Yes it's used `GServer.py:L617`.
How would do w/o using `reduce` since `pandas` does not have the functionality to join multiple DataFrames afaik ?

src/components/common/GProfiler/reducer.ts
```diff
@@ -0,0 +1,64 @@
...
+    display: false,
+    selectedTopGeneListsSizes: [],
+    isFetchingAvailableOrganisms: false,
+    availableOrganisms: [],
+    selectedOrganism: '',
+    selectedSortBy: '',
+    gProfilerToken: '',
+};
+
+const reducer: Reducer<State, GProfilerAction> = produce(
```
MaybeJustJames, line 19: This is an unfortunate thing about TypeScript, it's structurally typed so this is "correct". But `Reducer` is nominally incorrect (and confusing) here. [Here](https://github.com/reduxjs/redux/blob/master/src/types/reducers.ts#L29) is the redux `Reducer` type definition. If you want to properly type annotate this, be my guest (but with the redux `Reducer` type) or just don't type annotate the `reduce()` function.

src/components/common/GProfiler/reducer.ts
```diff
@@ -0,0 +1,64 @@
...
+    (draft: Draft<State>, action: GProfilerAction) => {
+        switch (action.type) {
+            case t.DISPLAY_MODAL:
+                draft.display = action.payload.display;
+                if (draft.display) {
+                    draft.selectedTopGeneListsSizes = [];
+                    draft.selectedOrganism = '';
+                    draft.selectedSortBy = '';
+                    draft.gProfilerToken = '';
+                    if (draft.availableOrganisms.length > 0) draft.error = '';
```
MaybeJustJames, line 29:
```suggestion
                    if (draft.availableOrganisms.length > 0) {
                        draft.error = '';
                    }
```

src/components/common/GProfiler/sagas.ts
```diff
@@ -0,0 +1,46 @@
+import { put, takeLatest, call } from 'redux-saga/effects';
+import * as R from 'ramda';
+
+import * as t from './actionTypes';
+import * as c from './constants';
+import { GProfilerOrganism } from './types';
+import * as Action from './actions';
+
+async function fetchJson(url: string) {
```
MaybeJustJames, line 9: This is a nice generic function! Could you put it in `src/api/` please?

src/components/common/GProfiler/utils.ts
```diff
@@ -0,0 +1,141 @@
+import {
+    GPROFILER_API_ENDPOINT__GENE_LIST_FUNCTIONAL_ENRICHMENT,
+    GPROFILER_LINK_MAX_LENGTH,
+} from './constants';
+
+import { FeatureMetadata, FeatureMetricTable } from './types';
+
+export const getMetricTable = (featureMetadata: FeatureMetadata) => {
```
MaybeJustJames, line 8: Does this return a `FeatureMetricTable`?

src/components/common/GProfiler/utils.ts
```diff
@@ -0,0 +1,141 @@
...
+    return featureMetadata.genes.map((gene: string, idx: number) => {
```
MaybeJustJames, line 9: This is a bit hairy...
Wondering how it could be simplified... could you operate on `R.zip(featureMetadata.genes, featureMetadata.metrics)` ? Maybe name the reducing function?

src/components/common/GProfiler/utils.ts
```diff
@@ -0,0 +1,141 @@
+import {
```
MaybeJustJames, line 1: `utils` is not a very descriptive filename. Maybe the link create and check functions could go in a `link.ts` and `getMetricTable`, `getAvailableSortBy` could go in `model.ts`?

src/components/common/GProfiler/utils.ts
```diff
@@ -0,0 +1,141 @@
...
+    selectedTopGeneListsSizes,
+    gProfilerToken,
+    selectedOrganism,
+    selectedSortBy,
+}: {
+    featureMetricTable: FeatureMetricTable;
+    selectedTopGeneListsSizes: number[];
+    gProfilerToken: string;
+    selectedOrganism: string;
+    selectedSortBy: string;
```
MaybeJustJames, line 105: Maybe name this interface

src/components/common/GProfiler/sagas.ts
```diff
@@ -0,0 +1,46 @@
...
+    return resp;
+}
+
+function* fetchAvailableOrganisms() {
+    yield put(Action.fetchingAvailableOrganisms(true));
+    const { data } = yield call(
+        fetchJson,
+        c.GPROFILER_API_ENDPOINT__AVAILABLE_ORGANISMS
+    );
+    if (!!data && !('message' in data)) {
```
MaybeJustJames, line 26: Could this bit of logic go somewhere else? Like `model.ts`?

MaybeJustJames: Cancel that. It's fine

src/components/common/GProfiler/sagas.ts
```diff
@@ -0,0 +1,46 @@
...
+        let data = await fetch(url);
+        resp = { data: await data.json() };
+    } catch (e) {
+        resp = { err: e.message };
+    }
+    return resp;
+}
+
+function* fetchAvailableOrganisms() {
+    yield put(Action.fetchingAvailableOrganisms(true));
```
MaybeJustJames, line 21: This should probably go in the reducer for `FETCH_AVAILABLE_ORGANISMS`. Doesn't need another action.

src/components/common/GProfiler/GProfilerModal.tsx
```diff
@@ -0,0 +1,299 @@
...
+        return [
+            getNumFeatures() < 100 ? getNumFeatures() : 100,
+            200,
+            300,
+            400,
+            500,
+        ].filter((topNumFeaturesValue) =>
+            topNumFeaturesValue <= getNumFeatures() ? true : false
+        );
+    };
```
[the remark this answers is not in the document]
dweemx: I'll put it in `model.ts`

src/components/common/GProfiler/TopGeneListsSelectionTable.tsx
```diff
@@ -0,0 +1,55 @@
...
+    selectedTopGeneListsSizes,
+    onSelectGeneList,
+}) => {
+    return (
+        <Table compact celled definition>
+            <Table.Header>
+                <Table.Row>
+                    <Table.HeaderCell />
+                    <Table.HeaderCell>
+                        Gene List with Number of Top Features to Use
```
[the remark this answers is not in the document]
dweemx: No, will change this to lowercase

src/components/common/GProfiler/reducer.ts
dweemx: Sorry I'm not following why this is incorrect
It states: 
```
 * @template S The type of state consumed and produced by this reducer.
 * @template A The type of actions the reducer can potentially respond to.
```
This what I provide and it should return a Reducer type no ?

src/components/common/GProfiler/sagas.ts
dweemx: [Here](https://github.com/reduxjs/redux/blob/master/src/types/reducers.ts#L29) they say: `* *Do not put API calls into reducers.*`

src/components/common/GProfiler/GProfilerModal.tsx
```diff
@@ -0,0 +1,299 @@
...
+    } = props;
+
+    const [availableSortBy, setAvailableSortBy] = useState<
+        { key: number; text: string; value: string }[]
+    >([]);
+    const [featureMetricTable, setFeatureMetricTable] = useState<
+        FeatureMetricTable
+    >([]);
+
+    useEffect(() => {
```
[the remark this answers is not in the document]
dweemx: Ok

src/components/common/GProfiler/reducer.ts
MaybeJustJames: Yeah this is the confusing part! React **and** Redux define a `Reducer` type. The reducer here is the Redux one, but you're importing/using the React one.

After I saw you typing this I went and found the correct type (https://github.com/aertslab/SCope/pull/381/files#diff-efbeaa2578aacd09fb3c964e4fbb2b4bR3) and used it in the cleanup PR.

src/components/common/GProfiler/sagas.ts
MaybeJustJames: Indeed. Maybe I'm misunderstanding but this isn't an API call, it's just flipping the fetching flag in the model to `true`?

src/components/common/GProfiler/reducer.ts
dweemx: Oh I see, thanks
I will import the `redux` one

src/components/common/GProfiler/sagas.ts
dweemx: Ok now I see what you mean
Yeah indeed it's better what you're suggesting

src/components/GProfiler/GProfilerModal.tsx
```diff
@@ -0,0 +1,265 @@
+import React, { useState, useEffect } from 'react';
```
MaybeJustJames, line 1: No need to import `useState` and `useEffect`

src/components/GProfiler/TopGeneListsSelectionTable.tsx
```diff
@@ -0,0 +1,55 @@
...
+                    );
+
+                    return (
+                        <Table.Row key={`tgls-${topGeneListsSize}`}>
+                            <Table.Cell collapsing>
+                                <Checkbox onChange={handleOnChangeGeneList} />
+                            </Table.Cell>
+                            <Table.Cell>
+                                {isSelected ? (
+                                    <b>{`Top ${topGeneListsSize}`}</b>
```
MaybeJustJames, line 40: `<b>` has semantics that make this usage strange? Do you mean to make this bold? In that case just `<hn>` and specify exact `font-weight` in CSS

MaybeJustJames (changes requested):
A couple of extremely minor changes, then rebase and we're good to merge.

dweemx:
> A couple of extremely minor changes, then rebase and we're good to merge.

Ok to skip the refactoring bits in `getMetricTable` ? I've tought about using `zip` but I don't think it's going to simplify dramatically the code

MaybeJustJames:
> > A couple of extremely minor changes, then rebase and we're good to merge.
> 
> Ok to skip the refactoring bits in `getMetricTable` ?

Oh yeah I forgot about that. I think that only needs the reducer to be a named function.

src/components/common/GProfiler/utils.ts
dweemx: Ok I named the reducing function

src/components/common/ViewerSidebar.jsx
```diff
@@ -1239,6 +1238,7 @@ class ViewerSidebar extends Component {
                             {markerTable}
                             {legendTable}
                             {downloadSubLoomButton()}
+                            <GProfilerModal featureMetadata={md} />
```
KrisDavie, line 1241: This still shows on the regulon tab but won't work because there is no sortable column. Two solutions:
1) Disable on regulon tab
2) Allow sorting by gene symbol if not other columns are available

KrisDavie (changes requested):
Just one last thing, other than that looks and works great (when the API works...)

src/components/common/ViewerSidebar.jsx
dweemx: I think 1. makes more sense

dweemx: Actually I will enable it only for Clustering result which is more generic

MaybeJustJames (approved):

KrisDavie (approved):
Awesome work!

MaybeJustJames (approved):

PR #372, opened 2020-08-05
Add a Result type and tests

### LINKED ISSUE #362: Principled error handling with a Result-like interface
The codebase currently has some error handling code returning tuple of the form:/
```python
return (True, "Success")
# ...or...
return (False, "Updated clusterings do not exist in file. Write error.")
```
This is similar to `Result` types in languages like [Rust](https://doc.rust-lang.org/std/result/) or `Either` in [Haskell](http://book.realworldhaskell.org/read/error-handling.html#errors.either). It would be great  to take this further and implement a custom `Result` type for SCope. I'm envisioning an interface of 4 functions. 2 `Result` creation functions:
```python
Result.failure("Some reason") # Returns (False, "Some reason")
Result.success() # Returns (True, "success")
```
And a matching function:
```python
Result.isFailure() # Returns the first element of the tuple
```

MaybeJustJames:
Closes #362

### DIFF
--- a/opt/scopeserver/result.py
+++ b/opt/scopeserver/result.py
@@ -44,7 +44,7 @@ class Result(Generic[T, E]):
 
     def is_ok(self) -> bool:
         """
-        Check is the computation succeeded.
+        Check if the computation succeeded.
 
         >>> ok(1).is_ok()
         True
--- a/opt/scopeserver/result.py
+++ b/opt/scopeserver/result.py
@@ -1,0 +1,115 @@
+# -*- coding: utf-8 -*-
+"""
+Generic error handling: represents 2 possibilities `ok`, or `err`.
+`ok` means the computation succeeded,
+`err` means the computation failed.
+
+There are 3 ways to construct a Result:
+  1. The `ok()` builder function to indicate success,
+  2. The `err()` builder function to indicate failure,
+  3. The `Result.from_optional()` method to build a result from an Optional
+"""
+
+from __future__ import annotations
+from typing import Generic, Optional, TypeVar, Callable
+
+# pylint: disable=invalid-name
+T = TypeVar("T")
+E = TypeVar("E")
+U = TypeVar("U")
+# pylint: enable=invalid-name
+
+
+class Result(Generic[T, E]):
+    """
+    The result of a computation that may fail.
+    This is a useful way to manage errors.
+    """
+
+    def __init__(self, value: Optional[T] = None, error: Optional[E] = None) -> None:
+        self._value: Optional[T] = value
+        self._error: Optional[E] = error
+
+    def is_ok(self) -> bool:
+        " Check is the computation succeeded. "
+        return self._value is not None
+
+    def is_err(self) -> bool:
+        " Check if the computation failed. "
+        return self._value is None
+
+    def map(self, transform: Callable[[T], U]) -> Result[U, E]:
+        """
+        Apply a function to a `Result`. If the result is `ok`, it will
+        be converted, otherwise the `err` will propogate.
+        """
+        if self._value is not None:
+            return Result(value=transform(self._value))
+
+        return Result(error=self._error)
+
+    def and_then(self, transform: Callable[[T], Result[U, E]]) -> Result[U, E]:
+        """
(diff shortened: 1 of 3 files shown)


MaybeJustJames:
> This is a bit broader than the original spec in #362, do you have examples of where the extra functions (`map`, `and_then`, `or_else` and `with_default`) would be used?

It's not really broader. Maybe you expected a different interface?
Beyond the tests, here are some more silly examples of how this `Result` could be used. I think I suggested using this type of `Result` in a comment on one of the active PR's but I can't find it now.

- **`with_default()`**: This is used to "extract" a success (`ok`) value. For example, if you have some function that may fail, and indicates that it may fail by returning `Result` then you can "extract" the `ok` with:
```python3
def function_may_fail(input: Int) -> Result[Int, Text]:
    if input < 0:
        return err("Only naturals")
   return ok(sqrt(input))

# answer will always contain a valid value: either some_default_value or the successful result of function_may_fail()
answer = function_may_fail(-1).with_default(0) # => 0 because the function failed
```
- **`map()`**: Gives you a mechanism to operate on `ok` values without worrying about if the `Result` is a success of failure. Imagine you have the `function_may_fail` as above, but you want to operate on its output. Say undo the `sqrt()` operation. Notice that `undo_sqrt()` does _not_ take a `Result` as input.
```python3
def undo_sqrt(input: Int) -> Int:
    return input**input

answer : Result[Int, Text] = function_may_fail(-1).map(undo_sqrt) # => err("Only naturals")
answer = function_may_fail(9).map(undo_sqrt) # => ok(9)
```
This means you can operate on `Result`s without caring that they're results.

- **`or_else()`**: Gives you an alternative. The alternative is executed if the initial computation fails. You can read it like "do this, _or else_ do this" or "do this and if it fails do this instead". Continuing the examples from above, let's introduce another function that can fail and use it as an alternative:
```python3
def possible_fail(text: Text) -> Result[Int, Text]:
    if text == "Only naturals":
        return err("An internal error occurred")
    return err("Some other error occurred")

answer : Result[Int, Text] = function_may_fail(-1).or_else(possible_fail) # => err("An internal error occurred")
answer = function_may_fail(9).or_else(possible_fail) # => ok(3)
```

- **`and_then()`**: Gives you a way to _compose_ or _chain_ operations together. You can read it as "do this, _and then_ do something else". Of course, this chain can be of any length: `ok(3).and_then(op1).and_then(op2)` etc. For example, again using the functions defined above:
```python3
def another_computation(input: Int) -> Result[Int, Text]:
    if input != 0:
        return ok(1 / input)
    return err("Undefined")

answer : Result[Int, Text] = function_may_fail(9).and_then(another_computation) # => ok(0.3333)
# or if function_may_fail fails:
answer = function_may_fail(-1).and_then(another_computation) # => err("Only naturals")
# or if another_computation fails:
answer = function_may_fail(0).and_then(another_computation) # => err("Undefined")
```

You will notice that in theses examples that I have **not** done any explicit error checking. That's the beauty of a `Result` monad.

If these descriptions/examples are helpful/useful I'll add them to the documentation.

MaybeJustJames:
I noted earlier that Python doesn't have pattern matching. One suggestion might be to emulate pattern matching with a `result()` function that takes 3 parameters:
1. what to do on `ok`,
1. what to do on `err`, and
1. the result to pattern match on.
Using such a function might look like this (I'm not entirely convinced this is a good idea or workable):
```python3
result(undo_sqrt, lambda: 0, function_may_fail(-1)) # => 0
result(undo_sqrt, lambda: 0, function_may_fail(9)) # => 9
```

opt/scopeserver/result.py
```diff
@@ -0,0 +1,115 @@
...
+    The result of a computation that may fail.
+    This is a useful way to manage errors.
+    """
+
+    def __init__(self, value: Optional[T] = None, error: Optional[E] = None) -> None:
+        self._value: Optional[T] = value
+        self._error: Optional[E] = error
+
+    def is_ok(self) -> bool:
+        " Check is the computation succeeded. "
```
MaybeJustJames, line 34:
```suggestion
        " Check if the computation succeeded. "
```

PR #376, opened 2020-08-13
Implement pre-commit hooks

### LINKED ISSUE #262: Automatic pre-commit hooks
Git pre-commit (pre-push) hooks are useful to avoid committing (pushing) code which is guaranteed to fail CI. The process of adding hooks for a given project is complicated and error prone... but can be automatied. Several tools already exist e.g.
* [Pre-commit](https://pre-commit.com/)
* [Husky](https://www.npmjs.com/package/husky)

KrisDavie:
Closes #262 

Take a look, try and try it out, see what you think

### DIFF
--- a/opt/pyproject.toml
+++ b/opt/pyproject.toml
@@ -51,6 +51,7 @@ requests = "^2.22.0"
 fastapi = "^0.54.1"
 hypercorn = "^0.9.5"
 methodtools = "^0.1.2"
+pre-commit = "^2.6.0"
 
 [tool.poetry.dev-dependencies]
 pytest = "*"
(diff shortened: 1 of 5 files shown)


opt/pyproject.toml
```diff
@@ -51,6 +51,7 @@ requests = "^2.22.0"
 fastapi = "^0.54.1"
 hypercorn = "^0.9.5"
 methodtools = "^0.1.2"
+pre-commit = "^2.6.0"
```
MaybeJustJames, line 54: Can you make this a dev-dependency please?

MaybeJustJames (changes requested):
Given this is an optional install I'm actually fine with it being a pre-commit hook as is. Just make `pre-commit` a dev dependency and I will approve.

KrisDavie:
To possibly be reconsidered in the future.

PR #383, opened 2020-09-21
Deploying SCope with Docker

dweemx:


docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+version: '3'
+
+services:
+    scope:
+        restart: always
+        image: bujesse/scope:1.0
```
MaybeJustJames, line 6: This should be a build

MaybeJustJames (changes requested):
Excellent! One minor point

docker-compose.yml
bujesse: Agreed, I'll make the update

docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+version: '3'
+
+services:
+    scope:
+        restart: always
+        image: scope:1.0
```
dweemx:
```suggestion
        build: .
```

README.md
```diff
@@ -116,3 +116,41 @@ Keys:
...
+    RewriteCond %{HTTP:Connection} upgrade [NC]
+    RewriteRule ^/?(.*) "ws://0.0.0.0:8080/$1" [P,L]
+```
+
+##### Example
+
+1. Copy `config.json` and modify with `"reverseProxyOn": true,` and `publicHostAddress` set to your domain
+
+1. ```SCOPE_CONFIG=/path/to/config.json SCOPE_PORT=8080 npm run build```
+
```
dweemx, line 151:
```suggestion
1. `docker-compose build`
```

docker-compose.yml
dweemx: I added a suggestion to this and also added one for the README

docker-compose.yml
dweemx: @bujesse, I think this change is also required to make `docker-compose build` work

docker-compose.yml
```diff
@@ -0,0 +1,26 @@
...
+        restart: always
+        image: bujesse/scope:1.0
+        expose:
+            - 55850
+            - 55851
+            - 55852
+            - 55853
+        volumes:
+            - static:/app
+        entrypoint: ["poetry", "run", "hypercorn", "--bind", "0.0.0.0", "main:scope_api"]
```
MaybeJustJames, line 14: NICE TO HAVE: HTTP requests on `$SCOPE_PORT` in the `/api/` namespace should be forwarded to the hypercorn application as well.

docker-compose.yml
bujesse: I made the update, this is much better.

bujesse:
Along with updating master, here are the latest changes. Please review and let me know what you think.

- build from dockerfile instead of pulling an image in the compose
- added configurable data directory to docker-compose to share data
  between host and dockerised data server
- removed the npm build step from the Dockerfile since the assets are
  served from the host
- cleaned up useless volume

Dockerfile
```diff
@@ -38,7 +38,10 @@ COPY ${SCOPE_CONFIG:-config.json} /app/config.json
```
dweemx, line 38: In the beginning of the Dockerfile:
```suggestion
FROM debian:latest

ARG SCOPE_PORT=8080
```

Dockerfile
```diff
@@ -38,7 +38,10 @@ COPY ${SCOPE_CONFIG:-config.json} /app/config.json
 
 # install the app
 RUN cd /app && npm install
-RUN cd /app && npm run build
+
+# No need to run the build when apache is serving assets from the host (see docker-compose.yml)
+# RUN cd /app && npm run build
+
 RUN cd /app/opt && poetry install
```
dweemx, line 46: To be added after:
```
# install the app
RUN cd /app && npm install
```

```suggestion
# build static html
RUN cd /app && SCOPE_CONFIG=./config.json SCOPE_PORT=${SCOPE_PORT} npm run build
```

README.md
```diff
@@ -149,7 +153,7 @@ Here is an example:
 
 1. ```SCOPE_CONFIG=/path/to/config.json SCOPE_PORT=8080 npm run build```
 1. `docker-compose build`
```
dweemx:
```suggestion
1. `docker-compose build --build-arg SCOPE_PORT=8080`
```

docker-compose.yml
```diff
@@ -3,14 +3,14 @@ version: '3'
...
             - 55850
             - 55851
             - 55852
             - 55853
         volumes:
-            - static:/app
+            - ${SCOPE_DATA_DIR:-./scope_data/}:/root/.local/share/SCope/
         entrypoint: ["poetry", "run", "hypercorn", "--bind", "0.0.0.0", "main:scope_api"]
     httpd:
         restart: always
```
dweemx: Would be nice to update the whole `docker-compose.yml` with this bits:
```suggestion
version: '3'

services:
    scope:
        restart: always
        build: .
        expose:
            - 55850
            - 55851
            - 55852
            - 55853
        volumes:
            - ${SCOPE_DATA_DIR:-./scope_data/}:/root/.local/share/SCope/
            - app-volume:/app
        entrypoint: ["poetry", "run", "hypercorn", "--bind", "0.0.0.0", "main:scope_api"]
    httpd:
        restart: always
        build: ./apache
        ports:
            - ${SCOPE_PORT:-80}:80
        depends_on:
            - scope
        volumes:
            - app-volume:/usr/local/apache2/htdocs/
volumes: 
    app-volume:
```

dweemx (commented):
Hi @bujesse,
I tested it locally and I had some issue starting SCope with `docker-compose`. So I made some changes in `Dockerfile`, `docker-compose` to make it work. I've added thoses changes as suggestions here.

docker-compose.yml
bujesse: The problem with this is that you're serving the assets from inside the container (sharing volumes between apache and  the scope app instead of the host). Personally, I think it's better to serve them from outside the container so you can run "npm run build" from outside when you need to change config.json. This is useful when you need to change the publicHostAddress without rebuilding the image.

Is there a reason you want to serve assets from inside the container?

Dockerfile
bujesse: I said this below, but I think the assets should be served from the host, unless you have a better reason. Otherwise you have to rebuild the image every time you change the config.

Dockerfile
bujesse: Another concern about building/serving the asset from the container is with this hardcoded argument you can't change the desired port you want to run the app on without changing this tracked file and rebuilding the image

docker-compose.yml
dweemx: Indeed having to rebuild the image each time you change you config is annoying.
What I was trying to avoid is having the `npm run build` run on different Node version than what the Docker image is using (v14).
I have Node `v9` on my computer. I followed the steps you wrote in the `README.md` and I wasn't not able to start SCope with the `docker-compose.yml` file.
I think this is because discrepancies between 2 node versions. This is why I suggested to run `npm run build` within the Docker container.
Maybe instead we could leave like that but update the `README.md` with and build the assets with the scope container ? What do you think ?

bujesse: Oh I see, I could see that being a problem. 

Are you thinking we should build the assets from within the container, but then have a shared volume with the host, so it still gets served from the host in the end? Or were you thinking to share a volume between scope and apache and have the assets built and served inside, like you had suggested?

dweemx: Yeah the former. I think your point of serving the assets from the host is a strong point so let's keep that.
But then instead of building the assets with the host, we could use the SCope docker instead in order to keep Node version consistent ? (So the `README.md` should be updated then)

bujesse: Got it, I'll make the change and test it

bujesse: Apparently you cannot copy files from a container in the Dockerfile, but you _can_ do it afterwards manually with the "docker cp" command. Would it be alright for me to add to the README that if you need to, you can grab the assets from within the container using that command?

bujesse: Or possibly even better; have 2 docker-compose files, one for serving from host, and one serving from container.

dweemx: Yes I think it's fine

bujesse: Please see the new changes and try again

dweemx: Just tested and it works for me when I follow the example to serve from container :+1:

docker-compose.yml
dweemx: @MaybeJustJames Could we defer this for later or do you have a suggestion for this ?

MaybeJustJames (changes requested):
Trying to run this from a fresh checkout I receive a constant flow of these errors:
```
scope_1  | TypeError: protobuf.loadProtoFile is not a function
scope_1  |     at Object.<anonymous> (/app/opt/scopeserver/bindserver/server.js:11:22)
scope_1  |     at Module._compile (internal/modules/cjs/loader.js:1076:30)
scope_1  |     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1097:10)
scope_1  |     at Module.load (internal/modules/cjs/loader.js:941:32)
scope_1  |     at Function.Module._load (internal/modules/cjs/loader.js:782:14)
scope_1  |     at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
scope_1  |     at internal/main/run_main_module.js:17:47
scope_1  | /app/opt/scopeserver/bindserver/server.js:11
scope_1  | gbBuilder = protobuf.loadProtoFile(path.join(__dirname,'grpc-bus.proto'));
```

MaybeJustJames:
Looks like I have to set `reverseProxyOn: true`.

dweemx:
> Trying to run this from a fresh checkout I receive a constant flow of these errors:
> 
> ```
> scope_1  | TypeError: protobuf.loadProtoFile is not a function
> scope_1  |     at Object.<anonymous> (/app/opt/scopeserver/bindserver/server.js:11:22)
> scope_1  |     at Module._compile (internal/modules/cjs/loader.js:1076:30)
> scope_1  |     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1097:10)
> scope_1  |     at Module.load (internal/modules/cjs/loader.js:941:32)
> scope_1  |     at Function.Module._load (internal/modules/cjs/loader.js:782:14)
> scope_1  |     at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
> scope_1  |     at internal/main/run_main_module.js:17:47
> scope_1  | /app/opt/scopeserver/bindserver/server.js:11
> scope_1  | gbBuilder = protobuf.loadProtoFile(path.join(__dirname,'grpc-bus.proto'));
> ```

You should follow `##### Example serve from container` section from the `README.md`

bujesse:
> Trying to run this from a fresh checkout I receive a constant flow of these errors:
> 
> ```
> scope_1  | TypeError: protobuf.loadProtoFile is not a function
> scope_1  |     at Object.<anonymous> (/app/opt/scopeserver/bindserver/server.js:11:22)
> scope_1  |     at Module._compile (internal/modules/cjs/loader.js:1076:30)
> scope_1  |     at Object.Module._extensions..js (internal/modules/cjs/loader.js:1097:10)
> scope_1  |     at Module.load (internal/modules/cjs/loader.js:941:32)
> scope_1  |     at Function.Module._load (internal/modules/cjs/loader.js:782:14)
> scope_1  |     at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
> scope_1  |     at internal/main/run_main_module.js:17:47
> scope_1  | /app/opt/scopeserver/bindserver/server.js:11
> scope_1  | gbBuilder = protobuf.loadProtoFile(path.join(__dirname,'grpc-bus.proto'));
> ```

I think you figured out that you have to set reverseProxyOn to true in the config.json before running. I suppose we could either make this the default in the config.json, or provide another file like config.docker.json that has the correct config in it and then have the default point there instead. Or we can leave it be. Let me know.

MaybeJustJames (approved):
Approved. Any fixes we need to make can be made later.

PR #407, opened 2020-12-10
Enable testing functions requiring a loom file

KrisDavie:
* Include "test" loom in repo, feature complete
* Enable testing of various loom based functions


opt/scopeserver/gen_test_loom.py
```diff
@@ -0,0 +1,256 @@
...
+from numpy.random import Generator, PCG64
+
+import zlib
+import json
+import base64
+import datetime
+from pathlib import Path
+
+
+def dfToNamedMatrix(df):
```
dweemx, line 13: Would be good to follow Python convention:
```suggestion
def df_to_named_matrix(df):
```
The function calls therefore should also be changed

opt/scopeserver/gen_test_loom.py
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    row_attrs = {
+        "Gene": gene_ids,
+        **embeddings_rows,
+        **clusterings_rows,
+        **regulons_rows,
+    }
+
+    metaJson = {
```
dweemx, line 238: I would also suggest to rename it
```suggestion
    meta_json = {
```

opt/test/test_loom.py
```diff
@@ -0,0 +1,145 @@
...
+    matrix, row_attrs, col_attrs, attrs = loom_file
+
+    # POC for custom data in the loom file.
+    num_cells = 100
+    _X = np.concatenate([rg.normal(n, 0.1, int(num_cells / 4)) for n in range(-2, 2)])
+    _Y = rg.normal(0, 0.1, num_cells)
+    main_embedding = pd.DataFrame(columns=["_X", "_Y"])
+    main_embedding["_X"] = _X
+    main_embedding["_Y"] = _Y
+    col_attrs["Embedding"] = Loom.dfToNamedMatrix(main_embedding)
```
dweemx, line 128: Shouldn't this be imported from `opt/scopeserver/gen_test_loom.py` ? Seems to be like redundant code

dweemx (changes requested):
Nice ! Just 1 small change requested about naming convention and another whether I'm not too sure of...

MaybeJustJames (approved):
This is great to have in, we can improve as needed.

opt/test/test_loom.py
KrisDavie: This actually generates different results but is also a proof of concept for generating data in the test to use in the loom file.

dweemx (approved):
:+1:

PR #410, opened 2021-01-18
Fix scope.aertlsab.org url to use https protocol

dweemx:


KrisDavie (approved):
Thanks Max!

dweemx:
And thanks @ghuls for reporting it!

PR #415, opened 2021-03-01
Search/Search space updates

KrisDavie:
Update terminal mappings for drosophila
    - Bump SS version as rebuild needed
Update order of search results
    - Correctly assign final categories to searches
    - Prefer genes/regulons over markers


dweemx (changes requested):
Great! I don't think it requires a new PR for package updates, only those 2 packages need to be updated I think:
```
npm install immer@8.0.1  to resolve 1 vulnerability
npm install --save-dev electron@11.3.0  to resolve 1 vulnerability
```
Or `electron` can even be uninstalled since we don't support it anymore ?

MaybeJustJames:
> Great! I don't think it requires a new PR for package updates, only those 2 packages need to be updated I think:
> 
> ```
> npm install immer@8.0.1  to resolve 1 vulnerability
> npm install --save-dev electron@11.3.0  to resolve 1 vulnerability
> ```
> 
> Or `electron` can even be uninstalled since we don't support it anymore ?

The bump to `immer` is a major version... are we sure there are no breaking changes that affect us?

opt/scopeserver/dataserver/utils/search.py
```diff
@@ -54,26 +54,39 @@ class ResultTypePair(NamedTuple):
 }
 
 
-def match_result_cost(term: str, result: str) -> Optional[int]:
+def match_result_cost(term: str, result: str, result_type: str) -> Tuple[int, Optional[int]]:
+    term_cost: Optional[int]
```
MaybeJustJames, line 58:
```suggestion
    term_cost: Optional[int] = None
```

And remove the `else` later

opt/scopeserver/dataserver/utils/search.py
```diff
@@ -54,26 +54,39 @@ class ResultTypePair(NamedTuple):
...
+    elif result.startswith(term) or result.endswith(term):
+        term_cost = 2
+    elif result.casefold().startswith(term.casefold()) or result.casefold().endswith(term.casefold()):
+        term_cost = 3
+    elif term in result:
+        term_cost = 4
+    elif term.casefold() in result.casefold():
+        term_cost = 5
+    else:
+        term_cost = None
```
MaybeJustJames, line 73:
```suggestion
```

opt/scopeserver/dataserver/utils/search.py
```diff
@@ -54,26 +54,39 @@ class ResultTypePair(NamedTuple):
 }
 
 
-def match_result_cost(term: str, result: str) -> Optional[int]:
+def match_result_cost(term: str, result: str, result_type: str) -> Tuple[int, Optional[int]]:
```
MaybeJustJames, line 57: This makes using the result more difficult. Could you sum these values? Would that be useful? Then you can just return an int.

opt/scopeserver/dataserver/utils/search.py
```diff
@@ -54,26 +54,39 @@ class ResultTypePair(NamedTuple):
...
 
 def sort_results(search_term: str, results: List[MatchResult]) -> List[MatchResult]:
-    costs = [(match_result_cost(search_term, match.element), match, result) for match, result in results]
+    costs = [
+        (match_result_cost(search_term, match.element, match.element_type), match, result) for match, result in results
+    ]
     sorted_costs = sorted(
-        [(cost, match, result) for cost, match, result in costs if cost is not None], key=lambda cost: cost[0]
+        [(cost, match, result) for cost, match, result in costs if cost[1] is not None],
+        key=lambda cost: (cost[0][0], cost[0][1]),
```
MaybeJustJames, line 89: This is where the complexity comes in. Lets avoid if possible.
```suggestion
        key=lambda cost: cost[0],
```

Could also use:
```python
import operator
operator.itemgetter(0)
```
rather than the lambda.

opt/scopeserver/dataserver/utils/search.py
```diff
@@ -167,16 +180,15 @@ def create_feature_description(
         if k[1] == "gene" and len(synonyms) > 0:
             descriptions[k].append(f"Synonyms: {', '.join(synonyms)}")
         elif k[1] in DEFINED_SEARCH_TYPES:
+            category_name = features[k]
+            category_type = feature_types[k]
+            desc_key = ResultTypePair(category_name, category_type)
+            features[desc_key] = features[k]
+            feature_types[desc_key] = feature_types[k]
             if DEFINED_SEARCH_TYPES[k[1]]["final_category"] == "cluster_category":
                 is_cluster = True
```
MaybeJustJames, line 189: Rather than the `if`/`else`, this can be reduced to:
```python
is_cluster = DEFINED_SEARCH_TYPES[k[1]]["final_category"] == "cluster_category"
```

package.json
```diff
@@ -114,8 +114,6 @@
         "babel-loader": "^8.1.0",
         "cross-env": "^5.1.4",
         "css-loader": "^3.5.3",
-        "electron": "^8.3.0",
```
MaybeJustJames, line 117: Lovely

MaybeJustJames (changes requested):
Looks mostly fine. I am worried about the additional complexity of the `match_result_cost()` signature. Can it be simplified to just return an `int` now?

dweemx (dismissed):
Nice work! Tested :heavy_check_mark:

MaybeJustJames (approved):
Great work!

dweemx (approved):

Issue #414, opened 2021-02-18
API migration


dweemx:
This issue is meant to track the migration of the current API (`gRPC` to the new one `FastAPI`)

PR #418, opened 2021-04-02
Feature/414 API migration - Part 1

dweemx:
Part 1 of #414 
Complete switch (`gRPC` implementation is removed) of the following endpoints to `REST` `FastAPI`:
- `getMyLooms`
- `getRemainingUUIDTime`
- `getUUID`


opt/main.py
```diff
@@ -3,33 +3,71 @@
...
+from scopeapi.model import GetRemainingUUIDTime, GetDatasets
 
 scope_api = FastAPI()
+scope_api.add_middleware(
+    CORSMiddleware,
+    allow_origins=["*"],
+    allow_credentials=True,
+    allow_methods=["*"],
+    allow_headers=["*"],
+)
```
MaybeJustJames, line 25: This is tricky. Maybe ok for now.

opt/main.py
```diff
@@ -3,33 +3,71 @@
...
+
 CONFIG = from_file(os.environ.get("SCOPE_CONFIG"))
 
+__scope_server = SCopeServer(config=CONFIG)
+__scope_server_legacy = LegacySCopeServer(config=CONFIG, new_server=__scope_server)
+__scope_api = SCopeAPI(server=__scope_server)
```
MaybeJustJames, line 31: Is there any reason you're using a dunder prefix here? I would prefer you avoid it.

opt/main.py
```diff
@@ -3,33 +3,71 @@
...
+
 
-@scope_api.get("/echo/{value}")
+@scope_api.get("/echo/{value}/")
```
MaybeJustJames, line 34: This endpoint can be removed.

MaybeJustJames (changes requested):
A general comment on organisation:

The top-level seperation of `scopeapi` and `scopeserver` is a mistake because there are guaranteed to be dependencies between them. Here's the organisation I would prefer to see:

```
.
├── alembic/
│   └── ...
├── alembic.ini
├── main.py
├── scopeserver
│   ├── api
│   │   ├── deps.py
│   │   ├── __init__.py
│   │   └── v1
│   │       ├── __init__.py
│   │       └── ...
│   ├── config.py
│   ├── database.py
│   ├── __init__.py
│   ├── models.py
│   └── schemas.py
└── tests
      └── ...
```

opt/main.py
dweemx: Yes did not find a better solution yet.

opt/main.py
dweemx: I will change that

PR #419, opened 2021-04-08
Fastapi organisation template

MaybeJustJames:
This introduces no-op functionality into the existing SCope application. No "wiring up" has been done.

Users are identified by an integer that the client must store in a Cookie. Projects have 2 alternative identifiers: an integer and a uuid-v4. Both of these are generated on the server. Each project stores its size and creation date so that a cron job can clean up as necessary. This means that clients no longer need to poll the server.

To run the server in _development_ configuration simply run the `./run-dev-env.sh` script. Note that the `--reload` option has been removed because it interacts badly with the legacy server. Until the legacy server is removed you will need to manually kill and restart the development server when you modify the code.


dweemx (dismissed):
Great work!
Would be nice to have kinda dev `README.md` to know how to start the server, setup database and migrations.

MaybeJustJames:
> Great work!
> Would be nice to have kinda dev `README.md`

Added. Also added a Postman collection

dweemx (approved):

MaybeJustJames:
I started to change "Project" to "Collection" when I realised that the name "Project" also works just as well for an "ad-hoc" collection of datasets as "Collection" does. It's just the concept that changes, rather than a "Project" being a rigid wall around a set of datasets its just an ad-hoc collection that can be.

KrisDavie (changes requested):
Other than the above comments, I'm also having issues stopping SCope. After running `run-dev-env.sh.sh`, a keyboard interrupt gives an `Aborted!` message, then a second or so later `Terminating Servers...` from SCope's logging. But the bind server never ends, I have needed to kill it manually before I can restart the server.

As noted on Slack also, h5py was updated when I ran `poetry install`, we need to lock below `3.0.0` as the way we search loom files breaks >=3.0.0.

config.json
```diff
@@ -12,6 +12,6 @@
     "orcidAPIClientID": "",
     "orcidAPIClientSecret": "",
     "orcidAPIRedirectURI": "",
-    "dataHashSecret": "",
-    "data": ""
+    "dataHashSecret": "0123456789abcdefghijklmnopqrstuvwxyz0123456789abcdefghijklmnopqr",
```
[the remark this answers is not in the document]
MaybeJustJames: Ok I'm fine with this. It will cause a validation error on an empty string so that fits nicely :)

opt/scopeserver/models.py
```diff
@@ -1,10 +1,54 @@
...
+    project = Column(Integer, ForeignKey("projects.id"), index=True)
+    user = Column(Integer, ForeignKey("users.id"), index=True)
+
+
+class User(Base):
+    " A SCope user. "
+    __tablename__ = "users"
+
+    id = Column(Integer, primary_key=True, index=True)
+    created = Column(DateTime, nullable=False)
```
dweemx, line 28: This should be filled automatically with a default value
```suggestion
    created = Column(DateTime, nullable=False, , default=datetime.datetime.utcnow)
```

opt/scopeserver/models.py
```diff
@@ -1,10 +1,54 @@
...
+
+
+class Project(Base):
+    " An ad-hoc collection of related datasets. "
+    __tablename__ = "projects"
+
+    id = Column(Integer, primary_key=True, index=True)
+    name = Column(String)
+    uuid = Column(String, index=True, nullable=False)
+    created = Column(DateTime, nullable=False)
```
dweemx, line 39: This should be filled automatically with a default value
```suggestion
    created = Column(DateTime, nullable=False, , default=datetime.datetime.utcnow)
```

opt/scopeserver/models.py
```diff
@@ -1,10 +1,54 @@
...
-class Test(Base):
-    __tablename__ = "test"
+class Dataset(Base):
+    " A single SCope dataset. "
+    __tablename__ = "datasets"
 
     id = Column(Integer, primary_key=True, index=True)
-    date = Column(Date)
+    name = Column(String)
+    created = Column(DateTime, nullable=False)
```
dweemx, line 51: This should be filled automatically with a default value
```suggestion
    created = Column(DateTime, nullable=False, , default=datetime.datetime.utcnow)
```

dweemx (approved):
Would just put a default value for the `created` columns in the database. Other than that I'm OK with the FastAPI organisation template :+1:

PR #421, opened 2021-04-27
Enable downloading a loom file of a selection of cells

KrisDavie:
This adds onto the previous functionality of downloading a cluster loom file with allowing a more specific selection using the lasso tool

### DIFF
--- a/src/components/common/ViewerSidebar.jsx
+++ b/src/components/common/ViewerSidebar.jsx
@@ -175,6 +175,139 @@ class ViewerSidebar extends Component {
             }
 
             return lassoSelections.map((lS, i) => {
+                let lassoDownloadSubLoomButton = () => {
+                    if (
+                        this.state.downloadSubLoomPercentage == null &&
+                        this.state.processSubLoomPercentage == null
+                    )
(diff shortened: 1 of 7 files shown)


src/components/common/ViewerSidebar.jsx
```diff
@@ -175,6 +175,139 @@ class ViewerSidebar extends Component {
             }
 
             return lassoSelections.map((lS, i) => {
+                let lassoDownloadSubLoomButton = () => {
```
MaybeJustJames, line 178: Please let's not continue doing this. That `<Button/>` should be a new component. And linting should tell you that `==` is forbidden? Use `===`

MaybeJustJames (changes requested):
And black

src/components/common/ViewerSidebar.jsx
dweemx: I'm OK to leave it like this for now since I already refactored all this file in an existing branch (`cleanup/373-frontend-stage4-viewersidebar`) that's ready

dweemx (approved):

PR #424, opened 2021-04-28
Fix incorrect truthy check failing on ID 0 coords

KrisDavie:


dweemx (approved):
This can probably be improved because logic behind is not easy to read imo. But for now is fine I think until we refactor `Compare.jsx`

src/components/pages/Compare.jsx
```diff
@@ -402,7 +402,10 @@ class Compare extends Component {
                                             superposition={superposition}
                                             activeCoordinates={
                                                 configuration == 'multi'
-                                                    ? multiCoordinates[j]
+                                                    ? multiCoordinates[j] !=
+                                                          null &&
+                                                      multiCoordinates[j] !=
+                                                          undefined
                                                         ? multiCoordinates[j]
                                                         : -1
```
MaybeJustJames, line 410: This is very very hard to parse. How about this:
```javascript
configuration === 'multi' ? R.defaultTo(-1, multiCoordinates[j]) : multiCoordinates[0]
```

MaybeJustJames (changes requested):

MaybeJustJames (approved):

PR #425, opened 2021-04-28
Allow more than 256 colours by looping colour constant list

KrisDavie:


opt/scopeserver/dataserver/utils/cell_color_by_features.py
```diff
@@ -140,23 +140,18 @@ def setAnnotationFeature(self, feature: str, annotations: Optional[List[Annotati
...
-        else:
-            raise ValueError("The annotation {0} has too many unique values.".format(feature))
-        # Set the reply
+        self.hex_vec = [constant.BIG_COLOR_LIST[i % len(constant.BIG_COLOR_LIST)] for i in ca_annotation_as_int]
+        colors = [constant.BIG_COLOR_LIST[i % len(constant.BIG_COLOR_LIST)] for i in range(len(md_annotation_values))]
 
         if annotations is not None:
             cellIndices = self.loom.get_anno_cells(annotations=annotations, logic=logic)
             self.hex_vec = np.array(self.hex_vec)[cellIndices]
+            colors = constant.BIG_COLOR_LIST[: len(md_annotation_values)]
```
MaybeJustJames, line 149: What if `md_annotation_values` is `> 256` too? It also seems to me that this should be a function:
```python3
from typing import Iterable, List
def toColours(index: Iterable[int]) -> List[str]:
  num_colours = len(constant.BIG_COLOR_LIST)
  return [constant.BIG_COLOR_LIST[i % num_colours] for i in index]
```

Then your change becomes:
```python3
self.hex_vec = toColours(ca_annotation_as_int)
colors = toColours(range(len(md_annotation_values)))
```
Note that this also makes your second assignment to `colors` (if `annotations is not None`) superfluous.
Just thinking about where this function should go... `constants.py` is probably not a bad idea.

opt/scopeserver/dataserver/utils/cell_color_by_features.py
```diff
@@ -140,23 +140,18 @@ def setAnnotationFeature(self, feature: str, annotations: Optional[List[Annotati
...
 
         reply = s_pb2.CellColorByFeaturesReply(
             color=self.hex_vec,
             vmax=self.v_max,
-            legend=s_pb2.ColorLegend(
-                values=md_annotation_values, colors=constant.BIG_COLOR_LIST[: len(md_annotation_values)]
-            ),
+            legend=s_pb2.ColorLegend(values=md_annotation_values, colors=colors),
```
MaybeJustJames, line 154: Hmm, why store a variable then? Why not justL
```suggestion
            legend=s_pb2.ColorLegend(values=md_annotation_values, colors=toColours(range(len(md_annotation_values)))),
```

opt/scopeserver/dataserver/utils/labels.py
```diff
@@ -38,13 +38,10 @@ def label_annotation(loom: Loom, embedding: int, feature: str) -> List[FeatureLa
     at the barycentre of the cell cluster.
     """
     values = loom.get_meta_data_annotation_by_name(name=feature)["values"]
-    all_annotations = loom.get_ca_attr_by_name(name=feature)
-    colours = uniq((constant.BIG_COLOR_LIST[values.index(i)] for i in all_annotations))
-
-    annotations = uniq(loom.get_ca_attr_by_name(name=feature))
+    colours = [constant.BIG_COLOR_LIST[i % len(constant.BIG_COLOR_LIST)] for i in range(len(values))]
```
MaybeJustJames, line 41:
```suggestion
    colours = toColours(range(len(values)))
```
Does this not break labels for clusters in the viewer?

MaybeJustJames (changes requested):

opt/scopeserver/dataserver/utils/constant.py
```diff
@@ -270,3 +271,8 @@ class Species(Enum):
     "69238c",
     "bf606c",
 ]
+
+
+def to_colours(index: Iterable[int]) -> List[str]:
```
MaybeJustJames, line 276: Add a docstring

```suggestion
def to_colours(index: Iterable[int]) -> List[str]:
    """ Convert indexes (`index`) into HTML hex colour values. 
         `index` can be a `range(some_number)` or any iterable collection of numbers.
    """
```

MaybeJustJames (changes requested):
I'm happy with it, just add a docustring to the new function and I will accept.

dweemx (approved):

opt/scopeserver/dataserver/utils/constant.py
KrisDavie: Good point. Done!

opt/scopeserver/dataserver/utils/labels.py
KrisDavie: This doesn't touch clusters, they have their own function because of needing to deal with `-1` values

MaybeJustJames (approved):

PR #460, opened 2021-07-02
Add the Eslint rule:  no-unused-vars

MaybeJustJames:
The change adds the ESLint rule no-unused-vars (https://eslint.org/docs/rules/no-unused-vars).
This was added because a common reviewer comment was to remove unused imports and variables,
this rule automates that review work.
This change also makes all existing client code pass this check.


MaybeJustJames:
@dweemx I cancelled the Server checks. They ran because I accidentally selected `master` as the target branch rather than `develop`. This PR doesn't touch server code.

dweemx (approved):

PR #469, opened 2021-07-22
Use the sanctuary Either type instead of the custom Result type

MaybeJustJames:
Sanctuary's `Either` is well tested and matches the Fantasyland spec the Ramda also conforms to which our `Result` did not. Therefore this is a good candidate to remove unnecessary code.
I do not expect the production bundle size to increase at all thanks to tree shaking.


KrisDavie (approved):
LGTM

MaybeJustJames:
The current release of Sanctuary depends on nodes `util` library. This is fixed on their `master` branch but is, as yet, unreleased. Therefore this PR is being closed until Sanctuary roles a release with this fixed.

PR #470, opened 2021-07-23
Remove pyscenic dependency

MaybeJustJames:
This change removes the pyscenic dependency and the gene set enrichment code that depended on it. This allows us to use Python 3.9 and drop Python 3.7 support.


README.md
```diff
@@ -26,7 +26,7 @@ Eventually the functionality from pySCENIC will be expanded and put in its own p
 
 ### Command Line
 
-You will need access to at least Python 3.7 do run this.
+You will need access to at least Python 3.8 do run this.
```
KrisDavie, line 29: Given we're updating the requirements here, I'm thinking it would be better to have a `Requirements` section and list `Python >=3.8` and `Nodejs >= 12(?)` there rather than here under the `Command Line` header, this isn't blocking the merge for me though.

KrisDavie (approved):
LGTM!

PR #474, opened 2021-07-27
Remove unused head injection code

MaybeJustJames:


MaybeJustJames:
Ignore for now until we implement proper auth

PR #477, opened 2021-07-29
orcid associated config keys should be upper-case

MaybeJustJames:


MaybeJustJames:
This PR is superseded by #479

PR #479, opened 2021-07-29
Simplify configuration

MaybeJustJames:
This makes the client and server independent of the global `config.json` file. This simplifies configuration by automating it into 2 configurations:
- development (run on localhost)
- production (run on scope.aertslab.org)
Further configurations can be added as necessary.
The major advantages of these changes are:
1. No more manually editing `config.json`
1. No more confusion between client config and server config


KrisDavie (changes requested):
This adds a lot of `scope.aertslab.org` specific code, but doesn't indicate to other users that these fields would need changing. I'd like a section in the README on how to run SCope on your own domain and which changes are neccessary for that.

KrisDavie (approved):

PR #493, opened 2021-09-15
Custom annotation colors #492

### LINKED ISSUE #492: Custom colors for annotations
It would be useful for a user to be able to set custom colors for a given annotation.

The idea would be to extend the JSON metadata with a new `colors` key:
```
{
   'name': 'batch', 
   'values': ['0', '1'], 
   'colors': ['#00FF00', '#800080']
}
```
Each color would be assigned to its respective annotation value (order matters). In this example this would be:
- `0` => `#00FF00`
- `1` => `#800080`

Loom files won't require to have this new `colors` key in order for SCope to work.

The result would look like:
image

dweemx:
Closes #492 

Not going to fix the client CI issue with outdated packages (this would be for another PR)


opt/scopeserver/dataserver/utils/constant.py
```diff
@@ -273,9 +274,17 @@ class Species(Enum):
...
+def to_colours(index: Iterable[int], color_list=None) -> List[str]:
     """Convert indexes (`index`) into pre-determined HTML hex colour values.
     `index` can be a `range(some_number)` or any iterable collection of numbers.
     """
-    num_colours = len(BIG_COLOR_LIST)
-    return [BIG_COLOR_LIST[i % num_colours] for i in index]
+    if color_list is not None and len(np.unique(index)) > len(color_list):
+        raise Exception(f"Not enough custom colors defined.")
+
+    COLOR_LIST = (
```
KrisDavie, line 284: I'm not a fan of using the same variable name just different in case, it could be a little confusing.

Maybe `final_color_list` or something similar is better?

opt/scopeserver/dataserver/utils/constant.py
```diff
@@ -273,9 +274,17 @@ class Species(Enum):
...
 
-def to_colours(index: Iterable[int]) -> List[str]:
+def to_colours(index: Iterable[int], color_list=None) -> List[str]:
     """Convert indexes (`index`) into pre-determined HTML hex colour values.
     `index` can be a `range(some_number)` or any iterable collection of numbers.
     """
-    num_colours = len(BIG_COLOR_LIST)
-    return [BIG_COLOR_LIST[i % num_colours] for i in index]
+    if color_list is not None and len(np.unique(index)) > len(color_list):
+        raise Exception(f"Not enough custom colors defined.")
```
KrisDavie, line 282: I would rather this report the error and then fallback to using the `BIG_COLOR_LIST`. This way, something is at least returned.

KrisDavie (changes requested):
Really nice feature. A couple comments, and can you please update `scope-json-metadata.schema.json` and `scope-json-metadata.test.json` with the new entries for custom colors.

opt/scopeserver/dataserver/utils/constant.py
dweemx: Fixed using @KrisDavie suggestion:
```
def to_colours(index: Iterable[int], color_list=None) -> List[str]:
    """Convert indexes (`index`) into pre-determined HTML hex colour values.
    `index` can be a `range(some_number)` or any iterable collection of numbers.
    """

    if color_list is None:
        final_color_list = BIG_COLOR_LIST
    elif len(np.unique(index)) > len(color_list):
        logger.warning(f"Not enough custom colors defined. Falling back to BIG_COLOR_LIST")
        final_color_list = BIG_COLOR_LIST
    else:
        final_color_list = [color[1:] if color.startswith("#") else color for color in color_list]

    num_colours = len(final_color_list)

    return [final_color_list[i % num_colours] for i in index]
```

KrisDavie (approved):

KrisDavie (approved):

PR #500, opened 2021-12-09
Client authentication

MaybeJustJames:
Integrates login to the client.
Also updates to react-routerv6, changes are as lazy as possible (so break a lot of features)


MaybeJustJames:
@KrisDavie note that the SemGrep issues weren't introduced in this PR

client/config/webpack.prod.js
```diff
@@ -45,7 +45,9 @@ module.exports = merge(common, {
                 client_id: 'APP-1QNL921F7P9FC3S4',
                 redirect_uri: 'https://scope.aertslab.org/',
             }),
-            API_PREFIX: 'https://scope.aertslab.org/api/v1/',
+            API_PREFIX: JSON.stringify('https://scope.aertslab.org/api/v1/'),
+            LOGIN_REDIRECT: JSON.stringify('/oidc_redirect'),
+            LOGOUT_REDIRECT: JSON.stringify('/'),
             __TEST_ONLY__: false,
```
kverstae: This should probably move to a `.env` file (or something similar).
Deployment on another domain would become easier this way + this will avoid accidentally leaking important API keys to the world.

client/src/api/fetch.ts
```diff
@@ -2,13 +2,22 @@ import { Result, error, success } from '../result';
...
+        if (typeof err === 'object') {
+            if (
+                err !== null &&
+                'statusText' in err &&
+                typeof (err as { statusText: unknown }).statusText === 'string'
+            ) {
+                return error((err as { statusText: string }).statusText);
+            }
+        }
+        return error(`Unknown error: ${JSON.stringify(err)}`);
```
kverstae: Almost exact duplication of `handleError()` in `api/err.ts` (https://github.com/aertslab/SCope/blob/8bfaffc27b8738c20a05d1b82cbd1b649ac9cd50/client/src/api/err.ts)

client/src/components/AppHeader.tsx
```diff
@@ -0,0 +1,105 @@
...
+    // TODO: Hacky implementation. To be refactored/reviewed properly
+    const countAnnotations = (metadata) => {
+        let count = 0;
+        for (const clustering of metadata['cellMetaData']['clusterings']) {
+            for (const cluster of clustering['clusters']) {
+                count += cluster['cell_type_annotation'].length;
+            }
+        }
+        return count;
+    };
```
kverstae: From what I can see, this method is only used the find out if there are any annotations. The actual number of annotations isn't needed for anything AFAIK.
If this assumption is correct, you could do something like this:
```
const hasAnnotations = (metadata) => {
    for (const clustering of metadata['cellMetaData']['clusterings']) {
        for (const cluster of clustering['clusters']) {
            if (cluster['cell_type_annotation'].length > 0) return true;
        }
    }
    return false;
}
```
This could save some iterations if you are lucky, worst case you have to loop everything again.

client/src/components/Viewer/reducer.ts
```diff
@@ -0,0 +1,20 @@
+import produce from 'immer';
+import { Reducer } from 'redux';
+
+import { State, initState } from './model';
+import { ViewerAction } from './actions';
+import * as AT from './actionTypes';
```
kverstae, line 6: Just a general comment. I noticed that the import of this module (and others) is not named consistently.
I have seen actionTypes been imported as `Action`, `t`, `AT`... which makes it somewhat confusing.
Not a major issue, but having a consistent naming of modules could improve readability (especially if you are not very familiar with the codebase :smile: )

client/src/redux/reducers/main.ts
```diff
@@ -4,10 +4,17 @@ import * as Action from '../actionTypes';
 import { MainState, MainAction, SESSION_READ } from '../types';
 
 const initialState: MainState = {
-    isAppLoading: true,
+    isAppLoading: false,
     uuid: '',
     sessionMode: SESSION_READ,
-    cookieConsent: false,
+    projects: [],
+    datasets: [],
```
kverstae, line 11: Would it make sense to add the datasets to the corresponding project instead of having 2 separate arrays?
This should make the link between dataset and project clearer instead of just having a property with the uuid of the project in the dataset objects.
This can of course only work if each dataset is guaranteed to be linked to a project, which I am not 100% certain...

kverstae (commented):
Changes make sense to me, didn't really look into detail at the implementation of things since I am not that familiar with how code is structured in this project.

There are 2 things that might be useful to add to the authentication as well:
1. Persisting authentication status. When you now refresh the page, authentication status is lost. Not really an issue since most users don't manually refresh there page, but would be nice if you could remain logged in after leaving the page.
2. Ability to log out. For now you can just refresh the page to do this, but a dedicated button would be nicer.

These changes are not necessarily needed in this PR. If they seem interesting enough, they can be added to the TODOs aswell

MaybeJustJames:
@kverstae Thanks for having a look at this! Your review is very helpful and insightful.
> There are 2 things that might be useful to add to the authentication as well:
> 1. Persisting authentication status. When you now refresh the page, authentication status is lost. Not really an issue since most users don't manually refresh there page, but would be nice if you could remain logged in after leaving the page. (#507)

I have some code ready to do this in a follow up PR.
> 2. Ability to log out. For now you can just refresh the page to do this, but a dedicated button would be nicer.

Indeed of course. That will be a TODO item. (#508)


> These changes are not necessarily needed in this PR. If they seem interesting enough, they can be added to the TODOs aswell

Indeed I'd like to avoid any further PR rot.

client/config/webpack.prod.js
MaybeJustJames: I disagree on 2 points.
1. This is the client build configuration. Using a `.env` mechanism just introduces unnecessary indirection. Users deploying to a different domain need only modify this configuration before building the JS bundle.
2. The client build knows nothing of any secrets. If it did the mere fact of distributing the JS would leak those secrets. Having them in a `.env` at build time does not fix that. Instead, the server knows the secrets and the client never has any knowledge of them. The trust model of OAuth2 auth-code flow is server-server meaning that the relying party (SCope) directly communicates with the identity provider to verify user credentials (code).

client/src/api/fetch.ts
MaybeJustJames: Good spot! Thanks

client/src/components/AppHeader.tsx
MaybeJustJames: You're right that this isn't great code (see the TODO) and that the `count` it computes is not usefully used. I am inclined to delete it outright because `pages/Annotations.tsx` should know how to handle `0` annotations without another component trying to hold its hand. @KrisDavie ?

client/src/components/Viewer/reducer.ts
MaybeJustJames: Yes. Agree.

client/src/redux/reducers/main.ts
MaybeJustJames: This is a good point. There are a few reasons I did it this way but I am open to alternatives since it's not entirely obvious what the access patterns will be like yet.
1. When viewing a dataset we don't really care which project it's part of so I don't want to have to index through a project to access a dataset
2. Projects are only really an organising and permissions mechanism for datasets. You can well imagine having a side-by-side view of 2 datasets from 2 different projects (so having indexing through projects becomes potentially awkward).
3. I suspect that indexing through projects will turn out to just be noise. We will rarely have to care which project a dataset is in (in the client) so requiring that we access a dataset through it's project, I suspect, will just add unnecessary typing.

client/config/webpack.prod.js
kverstae: 1. I am personally just not a big fan of having to change version controlled files when deploying something. 
2. Agreed, I was just trying to point to the ORCID and BITLY things in this config. However, these are (from what I can see) not used anywhere in the code anymore, so they can maybe just be removed? Secrets are indeed not very secret after bundling :smile:

client/src/components/AppHeader.tsx
kverstae: Having the menu item for annotation always be visible might also not be a bad thing for user experience.
It can become confusing if a menu item sometimes is visible and sometimes is not, without a clear indication of why this is.
Having the page itself handle missing annotations sounds indeed like a better approach.

client/src/redux/reducers/main.ts
kverstae: Makes sense to keep it like this then (for now). I was assuming that you could only have datasets of 1 project open at the same time, but this restriction doesn't really make any sense now that I think about it.

client/config/webpack.prod.js
MaybeJustJames: I've introduced an optional `SCOPE_SERVER_URL` env var for use at build time. If not set then defaults are used.

client/src/api/fetch.ts
MaybeJustJames: Done

client/src/components/AppHeader.tsx
MaybeJustJames: Ok I've made the button not depend on `hasAnnotations()`

client/src/components/Viewer/reducer.ts
MaybeJustJames: I won't fix in this PR but have filed an issue: #509

client/src/redux/sagas/scope/index.ts
```diff
@@ -75,7 +76,7 @@ export function* uploadFile(endpoint: string, project: string, file: File) {
 }
 
 export function* uploadRequest(action: T.UploadRequest) {
-    const url = new URL(API_PREFIX + 'project/dataset');
+    const url = new URL(SERVER_URL + API_PREFIX + 'project/dataset');
```
kverstae, line 79: Could this not use the API_URL from `api.ts`?

kverstae (approved):
Just 1 comment, otherwise good for me!

PR #510, opened 2022-03-09
Viewer redesign

### LINKED ISSUE #234: Replace JS WS server
This requires some investigation.
Experiment with:
* Apache/Nginx and WSGI
* Flask

MaybeJustJames:
I apologise for another massive PR! There are a few interesting changes here:

1. The "Gene" and "Compare" pages have been merged. They do **NOT** display anything yet (except for "Empty" or project UUID + dataset index).
2. A basic login modal is displayed when the application loads
3. After login, a button with the username is displayed in the top-right.
4. Adds a bunch of tests
5. Replaces Semantic-UI-CSS with maintained and compatible Formatic-UI-CSS
6. Remove the `PServer` (file upload server) as it's replaced with HTTP API facilities now.

Closes #234


client/src/components/AppHeader.tsx
```diff
@@ -1,59 +1,40 @@
...
             title: 'Regulon',
             icon: false,
         },
         {
-            display: metadata ? true : false,
             path: '/annotations',
             title: 'All Annotations',
             icon: false,
         },
         {
```
kverstae, line 32: Hiding all menu items (except home, tutorial and about) might be better UX.
It draws the attention more to the things that actually do something instead of some empty tabs

client/src/components/Main.tsx
```diff
@@ -49,8 +35,26 @@ export const Main: React.FC<{}> = () => {
         return <Navigate to={`/legacy/${session}`} replace />; // nosemgrep: typescript.react.security.audit.react-router-redirect.react-router-redirect
     }
 
+    const dispatch = useDispatch();
+
+    const keyevent = (event: KeyboardEvent): void => {
+        if (['Shift', 'Alt', 'Control'].includes(event.key)) {
```
kverstae: I am not a big fan of having 'Alt' as a modifier key. Having 'Shift' + 'Control' should be sufficient.
'Alt' already does things with the menu bar in most browsers and can cause some issues

client/src/components/Search/FeatureSearch.tsx
```diff
@@ -22,22 +21,21 @@ interface FeatureSearchProps {
...
+                gridColumn: '1 / 3',
+                gap: '10px',
+                padding: '10px 0px 0px 0px',
+            }}>
+            {[0, 1, 2].map((i) => (
+                <FeatureSearchBox
+                    field={`${props.identifier}-${i}`}
+                    filter={props.filter}
+                    colour={props.colour ? props.colour(i) : FEATURE_COLOURS[i]}
+                />
```
kverstae: A unique key is missing here

client/src/components/Viewer/ViewerWrapper.tsx
```diff
@@ -1,18 +1,169 @@
...
+                        return <div>Empty</div>;
+                    } else {
+                        return (
+                            <Placeholder
+                                dataset={viewer.dataset}
+                                project={viewer.project}
+                            />
+                        );
+                    }
+                }
```
kverstae: These should all have a unique key

client/src/components/ProjectsOverview.tsx
```diff
@@ -30,7 +30,7 @@ const ProjectView: React.FC<ProjectViewProps> = (props) => {
     const [displayUpload, setDisplayUpload] = useState(false);
```
kverstae, line 30: Adding a second dataset doesn't seem to work. This issue is not caused by changes in this PR.

client/src/components/Viewer/ViewerWrapper.tsx
```diff
@@ -1,18 +1,169 @@
...
+            })}
+
+            {state.remove ? (
```
kverstae: Something is not right with the tracking of the remove state.
Steps to reproduce: create a grid of 2x2 and remove the first row or column. The buttons at the sides should be green but they remain (inactive) red.

client/src/components/pages/Viewer.tsx
```diff
@@ -0,0 +1,30 @@
...
+                height: '100%',
+                padding: '0px 14px 0px 14px',
+            }}>
+            <Search.FeatureSearchGroup filter='all' identifier='viewer-page' />
+            <ViewerWrapper />
+            <RightSidebar
+                hideFeatures={false}
+                onActiveFeaturesChange={(_a, _b) => {
+                    return;
+                }}
```
kverstae, line 24: Maybe make this an optional prop? This empty arrow function doesn't really add anything valuable

kverstae (changes requested):
I like the idea of having all viewers in 1 place instead of split over multiple tabs.
It would be nice if these viewers could have variable dimensions instead of having to fit into a grid view (e.g 1 column x 2 rows, 3 columns x 1 row...). This will potentially add a lot of complexity, so not sure if this should be a priority

client/src/components/AppHeader.tsx
MaybeJustJames: I would prefer to keep this in the YAGNI folder. If this turns out to be true (way well be) then we can implement it. For now I'd like to keep it as simple as possible.

client/src/components/Main.tsx
MaybeJustJames: Done in 7d39bc95

client/src/components/Search/FeatureSearch.tsx
MaybeJustJames: Thanks. Fixed in 5397270a

client/src/components/Viewer/ViewerWrapper.tsx
MaybeJustJames: This is work-in-progress and is fine for now. Next step that @KrisDavie and I are working on is the viewer and this will be fixed then.

client/src/components/ProjectsOverview.tsx
MaybeJustJames: I cannot reproduce this. Can you be more detailed?

client/src/components/Viewer/ViewerWrapper.tsx
MaybeJustJames: Cannot reproduce. Is there some chance you let your finger off the 'Shift' button outside the browser window?

client/src/components/pages/Viewer.tsx
MaybeJustJames: The `RightSidebar` component isn't touched by this PR. This should be changed when the `RightSidebar` component is updated. I'd prefer to avoid PR creep as much as possible.

client/src/components/Viewer/ViewerWrapper.tsx
KrisDavie: For me, if I hold the shift key, they buttons flicker between red and green, sometimes remaining on red, sometimes on green.

client/src/components/ProjectsOverview.tsx
kverstae: I just try to add a second dataset but it won't show me the popup to upload the loom file and give the dataset a name.
I can replicate this in 2 ways (across 2 browsers):
1) create project 1 (OK) -> add dataset 1 (OK) -> add dataset 2 (no popup)
2) create project 1 (OK) -> add dataset 1 (OK) -> create project 2 (OK) -> add dataset to project 2 (no popup)

client/src/components/Viewer/ViewerWrapper.tsx
kverstae: This looks like a weird browser quirk.
In chrome everything just works as expected.
In firefox, the button just stays red after removing the first row/column even after releasing the shift key.
The state can be fixed by pressing ctrl, but this should not be necessary

MaybeJustJames: I've tested on Firefox 97.0, Firefox 78.15.0esr, Chromium 90.0.4430.212 (Developer Build), and Gnome Web 3.38.2. I cannot re-produce this. What browser(s) and browser version(s) are you using? Can you reproduce with all extensions disabled ([safe mode](https://support.mozilla.org/en-US/kb/troubleshoot-extensions-themes-to-fix-problems))?

kverstae: I tested with following browsers/versions:
Firefox 97.0.2 (Fedora 35): button stays red after removing first row/column and releasing shift
Firefox 98.0 (Fedora 35): button stays red after removing first row/column and releasing shift
Google Chrome 99.0.4844.51 (Fedora 35): works fine
Firefox 98.0.1 (Windows 10): buttons flicker. Releasing shift can either end in red or green
Google Chrome (Windows 10): buttons flicker again

I tried everything with and without my extensions enabled, but that didn't make any difference

PR #511, opened 2022-03-15
Implement Three.js viewer for redesigned viewers

KrisDavie:
Minimum working implementation of a three.js viewer which uses shaders to draw points.

Locally (with devtools closed!) I could display >15 million points and still reliably scroll around.

Some of this code is still _very_ hacky (project ID is hardcoded in `data.py` still.


MaybeJustJames:
Note that, because I had to manually adjust the time on my laptop, the most recent commit (223e729) appears to be the 3rd most recent

